PR #39, opened 2019-05-17
[No JIRA] Theme grey swatches

shaundon:
Quick-and-dirty colour swatches for the theming page. Will be useful for showing people our new colour theming support.

Kapture 2019-05-17 at 11 22 19

### DIFF
--- a/docs/src/components/ColorSwatch/ColorSwatch.js
+++ b/docs/src/components/ColorSwatch/ColorSwatch.js
@@ -30,20 +30,15 @@ const ColorSwatch = props => {
     backgroundImage: props.gradient,
   };
 
-  const classNames = [getClassName('bpkdocs-color-swatch')];
-
-  if (props.whiteColor) {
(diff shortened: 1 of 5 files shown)


docs/src/components/ColorSwatch/ColorSwatch.js
```diff
@@ -38,6 +38,9 @@ const ColorSwatch = props => {
   if (props.border) {
     classNames.push(getClassName('bpkdocs-color-swatch--border'));
   }
+  if (props.className) {
+    classNames.push(props.className);
+  }
```
ojcurtis: Can we potentially change this code to be more slick by doing something like this?

const classNames = getClassName(
    'bpkdocs-color-swatch',
    props.whiteColor && 'bpkdocs-color-swatch--light',
    props.border && 'bpkdocs-color-swatch--border',
    props.className,
  );

shaundon: Done

PR #50, opened 2019-06-11
No jira add floating action button page

bvitaliyg:


docs/src/constants/redirect-routes.js
```diff
@@ -25,6 +25,7 @@ export default {
   '/components/atoms/buttons': ROUTES.BUTTON,
   '/components/atoms/cards': ROUTES.CARD,
   '/components/atoms/chips': ROUTES.CHIP,
+  '/components/atoms/fab': ROUTES.FAB,
```
georgegillams: We don't actually need to add this - we only added redirect routes for existing things that people might have a bookmark for...

PR #89, opened 2019-08-19
Bpk 2708 web rating component

ojcurtis:

### DIFF
--- a/docs/src/pages/WebRatingPage/Rating.scss
+++ b/docs/src/pages/WebRatingPage/Rating.scss
@@ -16,26 +16,13 @@
  * limitations under the License.
  */
 
-/* @flow strict */
+@import '~bpk-mixins/index';
 
-import React from 'react';
+.bpk-docs-rating {
+  display: flex;
+  flex-wrap: wrap;
 
-import DocsPageWrapper from '../../components/DocsPageWrapper';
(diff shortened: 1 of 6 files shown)


docs/src/pages/WebRatingPage/Rating.scss
```diff
@@ -0,0 +1,28 @@
...
+ * limitations under the License.
+ */
+
+@import '~bpk-mixins/index';
+
+.bpk-docs-rating {
+  display: flex;
+  flex-wrap: wrap;
+
+  &-margin {
```
georgegillams: Should have 2 `-`s
Also can we name it something like `rating` rather than using a name that describes the effect?

ojcurtis: Sure thing! My only concern in calling it rating was due to the rating component itself having the same name so didn't want to mess the naming.

Inspiration was taken from `DialogExample.scss` 😄

georgegillams: At the end of the day this isn't shipped code, so no one really cares what it's called, but I would like 2 `-`s 😝

ojcurtis: Done! 😄

PR #130, opened 2019-09-27
[NO JIRA]: Adding new colours to the Colours docs page

ojcurtis:
colours

### DIFF
--- a/docs/src/components/ColorSwatch/ColorSwatch.js
+++ b/docs/src/components/ColorSwatch/ColorSwatch.js
@@ -28,6 +28,7 @@ const ColorSwatch = props => {
   const style = {
     backgroundColor: props.color,
     backgroundImage: props.gradient,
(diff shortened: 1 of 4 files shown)


docs/src/components/ColorSwatch/ColorSwatch.js
```diff
@@ -48,13 +49,15 @@ ColorSwatch.propTypes = {
   name: PropTypes.string.isRequired,
   className: PropTypes.string,
   color: PropTypes.string,
+  textColor: PropTypes.string,
```
georgegillams: It seems that `textColor` and `whiteColor` are both solving the same problem. Can we remove `whiteColor` from the swatch and replace all usages with `textColor={colorWhite}`?

ojcurtis: Sure why not 😄

PR #139, opened 2019-10-07
[BPK-3060]: Updated docs to use correct logo on homepage

ojcurtis:
The pride logo has also been corrected too.

Before:
Screenshot 2019-10-07 at 11 50 56

After:
Screenshot 2019-10-07 at 11 51 02


docs/src/helpers/logo-helper.js
```diff
@@ -17,10 +17,18 @@
  */
 
 import BackpackLogoWhite from '../static/backpack-logo-white.svg';
+import BackpackLogoGrey from '../static/backpack-logo-sky-grey.svg';
```
georgegillams: while we're at this, can we rename these to `backpack-logo-light` and `backpack-logo-dark`?

docs/src/helpers/logo-helper.js
```diff
@@ -17,10 +17,18 @@
...
 import BackpackLogoPride from '../static/backpack-logo-pride.svg';
+import BackpackLogoPrideGrey from '../static/backpack-logo-pride-sky-grey.svg';
 
-export const getBackpackLogo = () => {
+export const getBackpackLogo = (isHeroImage = false) => {
```
georgegillams: Can we rename this to be more `why` and less `what`?
```suggestion
export const getBackpackLogo = (lightVariant = false) => {
```

docs/src/helpers/logo-helper.js
ojcurtis: So these are in line with the spec from Will so will need to check as its not really based on light and dark settings more what page is shown on

docs/src/helpers/logo-helper.js
ojcurtis: I think `lightVariant` might be confusing as it doesn't depend on if the site is using a dark theme/mode more it is for solid vs coloured backgrounds

georgegillams: can we rename it `overImage` or something then, to capture that this is for use on top of images?

ojcurtis: Done!

PR #144, opened 2019-10-10
[BPK-3047]: Added StrategyPage to docs site

ojcurtis:
TODO:

- Use new images for page hero
- Tweak sizes to match design
- Update functionality to navigate to next page


docs/src/components/GuidelinesBuilder/GuidelinesPageBuilder.js
```diff
@@ -0,0 +1,169 @@
...
+  title: string,
+  subtitle: string,
+  link: string,
+};
+
+const Section = (props: SectionProps) => {
+  const { id, title, content, alternative } = props;
+  const sectionClassName = [
+    getClassName('bpk-docs-guidelines-page__section'),
+    alternative &&
```
tiagoengel: can we call this `alternate` instead of `alternative`. Alternate is more clear about what it does

ojcurtis: All done 😄

tiagoengel:
Looking good, there's something weird happening with the mobile version. (It's more visible if you toggle mobile view in chrome). The text is not breaking lines as it should, see:
Screenshot 2019-10-18 at 08 59 45

Another small thing is the text colour in `Next up`. Is a bit darker in the design:
Screenshot 2019-10-18 at 08 59 56
Screenshot 2019-10-18 at 09 02 38

docs/src/components/GuidelinesBuilder/GuidelinesPageBuilder.scss
```diff
@@ -0,0 +1,69 @@
...
+@import '~bpk-mixins';
+
+.bpk-docs-guidelines-page {
+  &__image {
+    height: $bpk-spacing-lg * 10;
+    color: $bpk-color-sky-gray-tint-01;
+  }
+
+  &__section {
+    padding: $bpk-spacing-xl * 2 $bpk-spacing-md * 10 $bpk-spacing-xs * 13 $bpk-spacing-xs * 10;
```
georgegillams:
```suggestion
    padding: $bpk-spacing-xl * 2 $bpk-spacing-xl * 5 $bpk-spacing-xs * 13 $bpk-spacing-lg * 2;
```

docs/src/components/GuidelinesBuilder/GuidelinesPageBuilder.scss
```diff
@@ -0,0 +1,69 @@
...
+    color: $bpk-color-sky-gray-tint-02;
+
+    @include bpk-breakpoint-mobile {
+      width: $bpk-spacing-xl * 20;
+    }
+  }
+
+  &__navigation {
+    display: flex;
+    padding: $bpk-spacing-xs * 10;
```
georgegillams:
```suggestion
    padding: $bpk-spacing-lg * 2;
```

docs/src/components/GuidelinesBuilder/GuidelinesPageBuilder.scss
```diff
@@ -0,0 +1,69 @@
...
+    padding-bottom: $bpk-spacing-xs;
+    justify-content: space-between;
+    background-color: $bpk-color-monteverde;
+    color: $bpk-color-white;
+    text-decoration: none;
+
+    &--icon {
+      padding-top: $bpk-spacing-xs * 4.7;
```
georgegillams: This seems like a very random value. We should stick to combinations of whole tokens...

docs/src/components/HeroSection/HeroSection.scss
```diff
@@ -33,6 +33,7 @@
   }
 
   &__heading {
+    z-index: 1;
```
georgegillams: Instead of doing this hack, could we place the `heading` element inside the `bpkdocs-hero-inner` div?

georgegillams: That would mean setting the background colour using `rgba` instead of opacity on the entire view.

docs/src/helpers/markdown-helper.js
```diff
@@ -0,0 +1,28 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+const getMarkdownString = readmeString =>
+  readmeString
```
georgegillams: Doesn't this file/logic already exist somewhere else?

docs/src/components/GuidelinesBuilder/GuidelinesPageBuilder.scss
ojcurtis: Yeah it didn't quite meet the design specs without it being this so 🤷‍♂

docs/src/helpers/markdown-helper.js
ojcurtis: Nope it was in the `InfoBuilderPage` before so moved it here for reusability

docs/src/components/HeroSection/HeroSection.scss
ojcurtis: I will try and see as remember we tried this yesterday and it didn't really work

docs/src/components/GuidelinesBuilder/GuidelinesPageBuilder.scss
georgegillams: In that case the design specs are wrong!!

docs/src/helpers/markdown-helper.js
ojcurtis: Have now migrated all usages to use this new helper util instead 😄

docs/src/components/GuidelinesBuilder/GuidelinesPageBuilder.scss
ojcurtis: Turns out I was wrong 🤦‍♂

docs/src/components/HeroSection/HeroSection.scss
ojcurtis: Going to leave it as it is right now as it solves the solution - but will revisit it soon as I hacked it around a lot when I removed this and it caused strange results

docs/src/components/NavigationFooter/NavigationFooter.js
```diff
@@ -0,0 +1,67 @@
...
+type NavigationFooterProps = {
+  id: string,
+  title: string,
+  link: string,
+};
+
+const NavigationFooter = (props: NavigationFooterProps) => {
+  const { id, title, link } = props;
+  return (
+    <a
```
tiagoengel: This needs to be a react router link. The way it is right now is refreshing to whole page instead of re-rendering the content inside.

ojcurtis:
> I just noticed another problem, the burger menu link is not showing up in the mobile view anymore.
> 
> Screenshot 2019-10-18 at 15 06 17

Ah yes this is a good spot! It will be to do with the opacity that we are applying to the div that it sits on. Will take a look 😄

ojcurtis:
I have now removed the setting of transparency of the images from the code as the correct assets to be provided will have this baked in so it will not cause any code hacking or breaking of the navigation bars 😄

PR #147, opened 2019-10-18
Fix small strategy page bugs

ojcurtis:


docs/src/components/NavigationFooter/NavigationFooter.scss
```diff
@@ -26,6 +26,11 @@
   color: $bpk-color-white;
   text-decoration: none;
 
+  @include bpk-breakpoint-mobile {
```
georgegillams: Why are we only doing this for mobile? Surely it should be vertically centred on desktop too?

ojcurtis: Not according to the designs

PR #149, opened 2019-10-23
Bpk 3048 add tone of voice page

tiagoengel:
- The close and tick are not exactly the same as the design but we don't have those ATM, so they will be updated in a different PR.

0 0 0 0_8080_guidelines_tone-of-voice (2)


docs/src/helpers/markdown-helper.js
```diff
@@ -20,9 +20,9 @@ const getMarkdownString = readmeString =>
   readmeString
     .replace(/^#.*$/m, '') // remove first h1
     .replace(/^>.*$/m, '') // remove first blockquote
-    .replace(/^#### /gm, '##### ') // replace h4 with h5
-    .replace(/^### /gm, '#### ') // replace h3 with h4
-    .replace(/^## /gm, '### ') // replace h2 with h3
-    .replace(/^# /gm, '## '); // replace h1 with h2
+    .replace(/^#### /m, '##### ') // replace h4 with h5
```
tiagoengel: this was not intentional, I'll revert 😅

docs/src/components/HeroSection/HeroSection.js
```diff
@@ -29,24 +30,28 @@ const getClassName = cssModules(STYLES);
...
+
   const classNames = [getClassName('bpkdocs-hero-section'), className];
   return (
     <div className={classNames.join(' ')} style={style}>
       <Heading
         level="h1"
         className={getClassName('bpkdocs-hero-section__heading')}
       >
-        {heading}
+        <BpkText textStyle="xxl" bold>
```
ojcurtis: Is this going to be applicable to all the new pages? We might be able to just pass the textStyle and bold into the Heading as i believe it uses BpkText under the hood?

docs/src/components/GuidelinesBuilder/GuidelinesPageBuilder.js
```diff
@@ -65,28 +83,39 @@ const Section = (props: SectionProps) => {
           level="h2"
           className={getClassName('bpk-docs-guidelines-page__heading')}
         >
-          {title}
+          <BpkText textStyle="xxl" bold>
```
ojcurtis: Is this going to be applicable to all the new pages? We might be able to just pass the textStyle and bold into the Heading as i believe it uses BpkText under the hood?

docs/src/components/HeroSection/HeroSection.js
tiagoengel: Yes is for all pages, but if you look at the design they all have this style so I don't see it as a problem

docs/src/components/GuidelinesBuilder/GuidelinesPageBuilder.js
tiagoengel: Yes all pages but that is the design so I don't see as a problem. I does't use BpkText, just style it using the mixins so passing `xxl` and `bold` won't work. I think this is the less `intrusive` way of doing this without affecting other pages

PR #155, opened 2019-10-25
Fixing guidelines icon size to be smaller

ojcurtis:
The icon was pointed out by Will to be bigger than the design, so we remove the `iconWidth` property to make it sized down to the correct size. 

This property was left over from the previous icon and implementation in the old site

Before 
Screenshot 2019-10-25 at 22 25 14

After
Screenshot 2019-10-25 at 22 28 21


georgegillams:
Cherry picked to a new branch: https://github.com/Skyscanner/backpack-docs/pull/157

PR #163, opened 2019-10-29
Move docs to npm

ojcurtis:
This PR now moves the docs dependencies away from using submodule file references for web components and will now pull the latest css packages from npm.

This improves reliability to use the latest packages that consumers use, test the css versions, and consume precompiled css for components aiding speed of startup.

A few modifications I have also made are:

- Removed `build-backpack` script as we are no longer consuming the local version - we therefore don't need to build it in CI or locally
- Removed the old `check-bpk-dependencies` as this is old from the monorepo
- Added a new script to use the `npx` for getting the css dependency updates


shaundon:
Should we be doing this for RN too? Fine if the answer is 'yes but in a later PR', I just wanna know the plan

ojcurtis:
> Could we remove `backpack` submodule please?

Let's do this is a follow up PR 😄

ojcurtis:
Some strange linting errors that exist on master are causing ci to fail so will fix as part of a separate PR

PR #175, opened 2019-11-11
Bpk 3053 add shadows page

ojcurtis:
Screenshot_2019-11-11 Shadows Backpack


tiagoengel:
Code looks good, I'll test locally.

I'm a bit worried with the size of the images. There is one with 1.7 mb for example and that is quite big... We also don't have any smaller variation for mobile with would make for a very bad experience in mobile networks.

PR #204, opened 2019-12-11
[BPK-3355] Add search function

shaundon:
Ended up being really simple. I've made it very straightforward, over-engineering it seemed pointless. It should suit our needs when we hook it up to UI anyway, we just need to pass in the things from `docs/src/layouts/links.js` and let it do its magic!

This is the library I'm using: https://fusejs.io/ It's well maintained and has good documentation so I think it'll be fine.


georgegillams:
So presumably we need to update `links.js` to contain all the relevant haystack info like keywords etc?

shaundon:
> So presumably we need to update links.js to contain all the relevant haystack info like keywords etc?

The only addition is the `keywords` part, and if it's omitted Fuse will just ignore it. So we could use this on `links.js` as it exists today and it should work fine. Then over time we can enrich it with keywords and the results will get more accurate.

PR #209, opened 2019-12-18
[BPK-3061] Add search to home page

shaundon:
- I moved the search logic I did last sprint into a `PageSearch` component.
- The `PageSearch` component doesn't have any styling information attached, **this is deliberate**, the styles comes from where it's used on `HomePage.js`. I think it's reasonable to assume we'll expose the search box to other pages later, so now the styles associated with making it look good on the home page are separate from the business logic.
- It's basic right now. This is also deliberate, I wanna deliver value now and then iterate on it later.
- In `docs/src/layouts/links.js`, we can add keywords and the search function will magically pick them up. I've added some to the checkbox entry as an example.

yo dawg we heard you like screenshots:

Screen Shot 2019-12-18 at 15 15 48
Screen Shot 2019-12-18 at 15 15 37
Screen Shot 2019-12-18 at 15 15 17


docs/src/layouts/links.js
```diff
@@ -80,6 +80,7 @@ const ComponentsLinks = [
         route: routes.CHECKBOX,
         children: 'Checkbox',
         tags: ['android', 'web'],
+        keywords: ['switch', 'toggle'],
```
shaundon: Example of adding keywords.

 Try typing 'toggle' and notice that checkbox is in the results 🙌

docs/src/layouts/links.js
```diff
@@ -97,6 +98,7 @@ const ComponentsLinks = [
         id: 'BLOCKQUOTE',
         route: routes.BLOCKQUOTE,
         children: 'Blockquote',
+        tags: ['web'],
```
shaundon: Missing, I added it.

docs/src/pages/HomePage/home-page.scss
```diff
@@ -131,14 +131,14 @@
 
   &__cards-container {
     display: flex;
-    margin: -$bpk-spacing-xl * 2 auto ($bpk-spacing-xl * 2 - $bpk-card-spacing-desktop);
+    margin: 0 auto ($bpk-spacing-xl * 2 - $bpk-card-spacing-desktop);
```
shaundon: This was the thing making the cards overlap the hero image. We don't need that anymore.

flow-typed/npm/react-router-dom_v4.x.x.js
```diff
@@ -0,0 +1,182 @@
+// flow-typed signature: 0bc486c8fc04d0bb37efa138223e4f67
+// flow-typed version: c6154227d1/react-router-dom_v4.x.x/flow_>=v0.104.x
+
+declare module "react-router-dom" {
```
shaundon: I added flow typings for `react-router-dom`.

(`npx flow-typed install react-router-dom@^4.3.1`)

PR #212, opened 2019-12-20
[No JIRA] Add search to sidebar

shaundon:
Added the search field to the sidebar! Design approved by Will 🎉 

Small convenience - press `Cmd` + `Ctrl` + `S` to focus the search field and scroll to the top of the page (so it's in view).

Very quick and dirty, it's just a power user feature. We'll probably use it more than anybody :P

### DIFF
--- a/docs/src/components/PageSearch/PageSearch.js
+++ b/docs/src/components/PageSearch/PageSearch.js
@@ -70,6 +70,10 @@ class PageSearch extends Component<Props, State> {
     document.addEventListener('keydown', this.handleKeyDown);
   }
 
+  componentWillUnmount() {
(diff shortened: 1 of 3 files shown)


docs/src/components/PageSearch/PageSearch.js
```diff
@@ -66,6 +66,10 @@ class PageSearch extends Component<Props, State> {
     this.state = { value: '', suggestions: [] };
   }
 
+  componentDidMount() {
+    document.addEventListener('keydown', this.handleKeyDown);
```
tiagoengel: we need to remove it in `componentWillUnmount `

shaundon: Good shout!

PR #214, opened 2019-12-23
[NO-JIRA] Fix navbar search box

tiagoengel:
This was breaking before when using the mouse to click in the suggestions because the `value` provided to the autosuggest input was null.

Kapture 2019-12-23 at 8 09 25


docs/src/components/PageSearch/PageSearch.js
```diff
@@ -125,7 +125,7 @@ class PageSearch extends Component<Props, State> {
     const inputProps = {
       id: 'search-pages',
       name: 'search-pages',
-      value,
+      value: value || '',
```
shaundon: OMG thank you for this!

PR #223, opened 2020-01-03
Fixing url to be https

ojcurtis:
Small fix


georgegillams:
Actually screw it - I've done it for ya

PR #231, opened 2020-01-16
[NO-JIRA] Fix ie11

tiagoengel:
This requires https://github.com/Skyscanner/backpack/pull/1767 to be merged first.


package.json
```diff
@@ -101,7 +101,7 @@
     "bpk-component-icon": "5.0.22-css.0",
     "bpk-component-image": "3.0.22-css.0",
     "bpk-component-infinite-scroll": "3.0.23-css.0",
-    "bpk-component-input": "5.0.22-css.0",
+    "bpk-component-input": "^5.0.24-css.0",
```
olliecurtis: Can you remove the caret from this? It causes npm to resolve to the non css version when this is present at the beginning due to how it views the latest and pre-release versions

package.json
```diff
@@ -135,7 +135,7 @@
     "bpk-component-tooltip": "4.0.21-css.0",
     "bpk-mixins": "19.0.21-css.0",
     "bpk-react-utils": "2.8.3-css.0",
-    "bpk-stylesheets": "4.1.2-css.0",
+    "bpk-stylesheets": "^4.1.4-css.0",
```
olliecurtis: Same as above

tiagoengel: A true, I tried to force that, not sure what happened

PR #357, opened 2020-02-03
[BPK-3437]: Typescale rollout

olliecurtis:
Updated to new dependency versions
Updated `tokens-helper.js` to add support for the larger sizes available on web

Typesetting


docs/src/helpers/tokens-helper.js
```diff
@@ -99,6 +99,15 @@ export const getTokens = (tokens, keys = null) =>
       if (token.name.indexOf('_XXL') !== -1) {
         return 7;
       }
+      if (token.name.indexOf('_XXXL') !== -1) {
```
georgegillams: We should have a helper function somewhere for this. Something like `includes(token.name, '_XXXL')`. But that's a separate issue!

PR #383, opened 2020-03-11
[BPK-3730] Add usage guidelines for picker component

shaundon:
0 0 0 0_8080_components_picker_platform=native

### DIFF
--- a/docs/src/pages/NativePickerPage/NativePickerPage.js
+++ b/docs/src/pages/NativePickerPage/NativePickerPage.js
@@ -29,8 +29,6 @@ const components = [
   {
     id: 'default',
     title: 'Default',
-    blurb:
-      'Picker trigger elements which display the currently selected value and can be tapped to present the actual picker.',
     screenshots: [
       {
         title: 'iOS',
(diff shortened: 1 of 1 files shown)


docs/src/pages/NativePickerPage/NativePickerPage.js
```diff
@@ -29,8 +29,6 @@ const components = [
   {
     id: 'default',
     title: 'Default',
-    blurb:
-      'Picker trigger elements which display the currently selected value and can be tapped to present the actual picker.',
```
shaundon: This sentence didn't even make sense so I just removed it. It's self evident anyway.

docs/src/pages/NativePickerPage/NativePickerPage.js
```diff
@@ -65,6 +63,12 @@ const NativePickerSubPage = ({ ...rest }) => (
     components={components}
     readme={readme}
     showMenu
+    usageTable={{
+      dos: ['Use for short lists of options.'],
+      donts: [
+        'Use for lots of options where the user may need to scroll a lot, for example a list of countries. In this situation, consider using a FlatList.',
```
olliecurtis: Might be worth linking to the FlatList component here?

shaundon: It only accepts strings, so I thought let's just fix that later, seeing as we'll probably move to a 'proper' guidelines thing soon that's more advanced than this. But good spot!

PR #385, opened 2020-03-11
[BPK-3734]: Add design tab for icons

olliecurtis:
## **Background**
The purpose of this PR is to add the ability for a design tab for the icons page to hold the list of icons that consumers can use.

## **Changes:**

- DocsPageWrapper.js - now includes a platform query `design` this will **future proof** us should we want to add design tabs for components or other pages in the future. By default the page will try to select the `design` tab first but if not available will navigate the correct page for the component
- Added `IconsDesignPage` that now holds the icon list and search functionality fields
- Removed icon list and search field from `WebIconsPage` as this is no longer required here.

## **Screenshot:**
IconsPage

### DIFF
--- a/docs/src/components/DocsPageWrapper/DocsPageWrapper.js
+++ b/docs/src/components/DocsPageWrapper/DocsPageWrapper.js
@@ -37,14 +37,16 @@ import STYLES from './DocsPageWrapper.scss';
 const getClassName = cssModules(STYLES);
 
 const contentShape = PropTypes.oneOfType([PropTypes.string, PropTypes.node]);
-const platformQueryParamRegex = /platform=(android|ios|native|web)/;
+const platformQueryParamRegex = /platform=(design|android|ios|native|web)/;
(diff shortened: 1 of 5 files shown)


docs/src/components/DocsPageWrapper/DocsPageWrapper.js
```diff
@@ -54,6 +56,14 @@ const PlatformNav = ({
     autoScrollToSelected
     className={getClassName('bpkdocs-page-wrapper__platform-switcher')}
   >
+    <BpkHorizontalNavItem
+      name="design"
+      disabled={disabledDesignTab}
```
shaundon: It's `disable`, not `disabled` on the other platforms

```suggestion
      disabled={disableDesignTab}
```

olliecurtis: Good spot all fixed now

PR #387, opened 2020-03-16
[BPK-3750] Add placeholder page for design tab

shaundon:
Unfortunately the linter changed a lot of these files, so I'll list which ones need looking at:

* `DocsPageWrapper.js`
* `DesignPlaceholderPage.js`

I tweaked the logic so that design is 'last' in the hierarchy of pages to automatically show users. I think this is the right decision as it means users won't see that page unless they explicitly go to it, which seems more useful. We don't want to be in people's face, just have it there as a small thing.

I made up the text we show - very happy to change it if people have a better suggestion. I wasn't sure what to put.

Screen Shot 2020-03-16 at 13 26 29

### DIFF
--- a/docs/src/components/DocsPageWrapper/DocsPageWrapper.js
+++ b/docs/src/components/DocsPageWrapper/DocsPageWrapper.js
@@ -30,6 +30,7 @@ import {
   getPlatformFromLocalStorage,
 } from '../../helpers/platform-helper';
 import Heading from '../Heading';
+import DesignPlaceholderPage from '../../pages/DesignPlaceholderPage';
 
 import Blurb from './Blurb';
 import STYLES from './DocsPageWrapper.scss';
@@ -148,16 +149,16 @@ const DocsPageWrapper = props => {
   const platformFromLocalStorage = getPlatformFromLocalStorage();
   if (canUsePlatformPreference(platformFromLocalStorage)) {
     initiallySelectedPlatform = platformFromLocalStorage;
-  } else if (designSubpage) {
-    initiallySelectedPlatform = 'design';
+  } else if (webSubpage) {
+    initiallySelectedPlatform = 'web';
   } else if (androidSubpage) {
     initiallySelectedPlatform = 'android';
   } else if (iosSubpage) {
     initiallySelectedPlatform = 'ios';
   } else if (nativeSubpage) {
     initiallySelectedPlatform = 'native';
-  } else if (webSubpage) {
-    initiallySelectedPlatform = 'web';
+  } else if (designSubpage) {
+    initiallySelectedPlatform = 'design';
   }
 
   const platformQueryParamMatches = platformQueryParamRegex.exec(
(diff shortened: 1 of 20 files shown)


docs/src/components/DocsPageWrapper/DocsPageWrapper.js
```diff
@@ -227,7 +228,7 @@ DocsPageWrapper.propTypes = {
 
 DocsPageWrapper.defaultProps = {
   blurb: null,
-  designSubpage: null,
+  designSubpage: <DesignPlaceholderPage wrapped />,
```
k0nserv: Hmm won't this only be evaluated a single time which means it will be the same instance in each case? I feel like that would be bad and will confuse the reconciler

docs/src/pages/DesignPlaceholderPage/DesignPlaceholderPage.js
```diff
@@ -0,0 +1,37 @@
...
+ */
+
+import React from 'react';
+
+import DocsPageBuilder from '../../components/DocsPageBuilder';
+import IntroBlurb from '../../components/IntroBlurb';
+import Paragraph from '../../components/Paragraph';
+
+const blurb = [
+  <IntroBlurb>Design guidelines needed!</IntroBlurb>,
```
olliecurtis: Could we maybe have something like?
```suggestion
  <IntroBlurb>Design guidelines do not currently exist for this component, but you can help!</IntroBlurb>,
```

docs/src/components/DocsPageWrapper/DocsPageWrapper.js
shaundon: Ah really? It just seemed like the most sensible, React way to do it?

docs/src/pages/DesignPlaceholderPage/DesignPlaceholderPage.js
olliecurtis: Or something more than `Design guidelines needed`

docs/src/components/DocsPageWrapper/DocsPageWrapper.js
k0nserv: Yeah, but I wonder if it's a good idea

```
DocsPageWrapper.defaultProps = {
  designSubpage: React.createElement(DesignPlaceholderPage, { wrapped: true }, null),
};
```

Which means that every `DocsPageWrapper` will have the exact same instance of the element. I'm unsure if React relies on the fact that each element instance will only exist in a single place in the view hierarchy

shaundon: Just been chatting to George and we're going to do it a different way than this anyway, so I'll resolve your comment

docs/src/pages/DesignPlaceholderPage/DesignPlaceholderPage.js
shaundon: Yeah I like it 😄

PR #425, opened 2020-04-06
[BPK-3801] Update tooltips page to add an example of dark tooltips

shaundon:
Added example as well as some usage advice to prevent misuse.

Screen Shot 2020-04-06 at 08 51 57

Screen Shot 2020-04-06 at 08 52 02


docs/src/pages/TooltipsPage/TooltipsPage.js
```diff
@@ -58,6 +58,27 @@ const components = [
...
+    examples: [
+      <BpkTooltip
+        id="my-tooltip"
+        type={TOOLTIP_TYPES.dark}
+        target={
+          <Heading
+            level="h3"
+            className={getClassName('bpkdocs-tooltips-page__heading')}
+          >
+            DAR
```
olliecurtis, line 74: Should this say Dark or LHR?

shaundon: It's an airport code :P The other examples said LHR and EDI, so I added a third one

shaundon: It's just the last place I went https://en.wikipedia.org/wiki/Dar_es_Salaam

olliecurtis: Ah ok, can you update the hover line to change to this airport instead of London Heathrow, just for consistency?

PR #431, opened 2020-04-07
[NO-JIRA] Fix typesetting info

georgegillams:

### DIFF
--- a/docs/src/pages/TypesettingPage/TypesettingPage.js
+++ b/docs/src/pages/TypesettingPage/TypesettingPage.js
@@ -54,11 +54,12 @@ const components = [
       ios: (
         <section>
           <Paragraph>
-            When designing for iOS, use <strong>SF Pro Text</strong>.
+            When designing for iOS, use <strong>Skyscanner Relative iOS</strong>
+            .
           </Paragraph>
           <BpkList>
-            <BpkListItem>Regular</BpkListItem>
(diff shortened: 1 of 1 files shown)


docs/src/pages/TypesettingPage/TypesettingPage.js
```diff
@@ -68,18 +69,22 @@ const components = [
       android: (
         <section>
           <Paragraph>
-            When designing for Android, use <strong>Roboto</strong>.
+            When designing for Android, use{' '}
```
olliecurtis, line 72: `{' '}` - this might be accidental?

georgegillams: prettier has added that to maintain the space.
If we write 
```
...for Android, use
<strong>Skyscanner...
```
Then it will render ...for Android, use**Skyscanner...**

The `{' '}` tells it to render a space before the next element, so it will be ...for Android, use **Skyscanner...**

olliecurtis: Hmm ok as it hasn't done that for the iOS and Web version 🤔

georgegillams: Because the line length was such that it didn't need to split it. Must be the extra 4 letters in Android

olliecurtis: 🤷‍♂ who knows

PR #453, opened 2020-04-22
[NO-JIRA] Fix snackbar metadata

georgegillams:


docs/src/layouts/links.js
```diff
@@ -98,7 +98,7 @@ const ComponentsLinks = [
         id: 'SNACKBAR',
         route: routes.SNACKBAR,
         children: 'Snackbar',
-        tags: ['android'],
+        tags: ['android', 'native'],
```
k0nserv, line 101: We should document this for iOS too

georgegillams: True, but that's not what this PR is about

PR #473, opened 2020-05-06
[BPK-3915]: Added new Indeterminate checkbox example

olliecurtis:
Screenshot 2020-05-06 at 12 39 21


docs/src/pages/WebCheckboxPage/WebCheckboxPage.js
```diff
@@ -53,6 +54,36 @@ const components = [
       </form>,
     ],
   },
+  {
+    id: 'indeterminate',
+    title: 'Indeterminate',
+    blurb: [
+      <Paragraph>
+        The indeterminate flag is only a visual clue, it does not affect the
```
olliecurtis, line 62: This is taken from the README and trimmed a line just to make this option clear

shaundon:
I think the example should be more true-to-life to illustrate better what this style is for. How about something like:

```
[-] Fruits
    [x] Apple
    [ ] Banana
    [ ] Quince 
```

docs/src/pages/WebCheckboxPage/WebCheckboxPage.js
```diff
@@ -53,6 +54,36 @@ const components = [
...
+        checked state of the checkbox. This is used when there is checkbox group
+        and the parent displays this state when not all children checkboxes are
+        selected.
```
shaundon, line 65:
```suggestion
        The indeterminate prop is only a visual clue, it does not affect the
        checked state of the checkbox. This is used when there is a checkbox group
        and the parent displays this state when not all child checkboxes are
        selected.
```

olliecurtis: Sounds good, will need the README in main Backpack to be updated too as this comes from there

shaundon: Ah yeah - you can probably just push that to master as it's implicitly been reviewed as part of this PR.

olliecurtis: Addressed

PR #517, opened 2020-05-27
[BPK-2527]: Added text link component docs

olliecurtis:
Link Page

### DIFF
--- a/docs/src/layouts/links.js
+++ b/docs/src/layouts/links.js
@@ -56,7 +56,7 @@ const ComponentsLinks = [
         id: 'LINK',
         route: routes.LINK,
         children: 'Link',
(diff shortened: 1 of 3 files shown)


.gitmodules
```diff
@@ -1,16 +1,16 @@
...
 [submodule "backpack-react-native"]
 	path = backpack-react-native
 	url = https://github.com/Skyscanner/backpack-react-native
-	##ignore = all
+	###ignore = all
 [submodule "backpack"]
 	path = backpack
 	url = https://github.com/Skyscanner/backpack
-	##ignore = all
+	###ignore = all
```
shaundon, line 16: This shouldn't happen but on the other hand it's not affecting anything so oh well

PR #552, opened 2020-06-26
[No Jira] Update fuse dependency and fix breaking change

shaundon:
Closes #551


shaundon:
Approving my own PR like a boss

PR #593, opened 2020-07-09
[BPK-4068]: New checkbox and radio design doc updates

olliecurtis:
Checkbox

Radio


docs/src/pages/WebCheckboxPage/WebCheckboxPage.js
```diff
@@ -141,6 +141,31 @@ const components = [
       </form>,
     ],
   },
+  {
+    id: 'light',
+    title: 'Light',
```
shaundon, line 146: Should we call this `white` to match the prop name?

```suggestion
    title: 'White',
```

docs/src/pages/WebCheckboxPage/WebCheckboxPage.js
```diff
@@ -141,6 +141,31 @@ const components = [
       </form>,
     ],
   },
+  {
+    id: 'white',
+    title: 'Light',
+    dark: true,
+    examples: [
```
shaundon, line 148: Can you add a blurb prop to this too, just to describe what it does as I think it's not the most obvious (that it only changes the label, not the actual checkbox).

```suggestion
    blurb: 'You can change the label to white for displaying on dark backgrounds.',
    dark: true,
    examples: [
```

(Same comment for radio too).

docs/src/pages/WebCheckboxPage/WebCheckboxPage.js
olliecurtis: I was thinking that - but also thinking about inclusive language 🤔

docs/src/pages/WebCheckboxPage/WebCheckboxPage.js
olliecurtis: Sounds good!

docs/src/pages/WebCheckboxPage/WebCheckboxPage.js
shaundon: Ahh, yeah. I think we're literally describing a colour though so it's ok — it's not like whitelist vs allowlist

PR #600, opened 2020-07-15
[BPK-4093]: Fixed radio spacing

olliecurtis:
Added 2px margin on both top and bottom of the radio buttons

Current:
Screenshot 2020-07-15 at 12 44 09

New:
Screenshot 2020-07-15 at 12 38 47

### DIFF
--- a/docs/src/layouts/SideNavLayout/NavListFilter.scss
+++ b/docs/src/layouts/SideNavLayout/NavListFilter.scss
@@ -19,9 +19,18 @@
 
 .bpkdocs-nav-list-filter {
   &__option {
+    margin-top: $bpk-spacing-xs / 3;
+    margin-bottom: $bpk-spacing-xs / 3;
+
     @include bpk-margin-trailing($bpk-spacing-sm);
 
-    :last-child {
+    &:first-child {
+      margin-top: 0;
+    }
+
+    &:last-child {
+      margin-bottom: 0;
+
       @include bpk-margin-trailing(0);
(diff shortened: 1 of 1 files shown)


docs/src/layouts/SideNavLayout/NavListFilter.scss
```diff
@@ -19,6 +19,9 @@
 
 .bpkdocs-nav-list-filter {
   &__option {
+    margin-top: $bpk-spacing-xs / 3;
+    margin-bottom: $bpk-spacing-xs / 3;
+
```
georgegillams, line 24:
```suggestion
    margin-bottom: $bpk-spacing-sm / 3;

    &:last {
      margin-bottom: 0;
    }
```

olliecurtis: Maybe the `&` is the thing missing here as I tried updating to do that but everything in the current `:last-child` was having no affect

olliecurtis: I also added the top margin so that it made it a lot more balanced as just doing the bottom margin made it off balance 😄

olliecurtis: Turns out it was the missing `&` that caused it so fixed it to actually work now

georgegillams: 👍

PR #601, opened 2020-07-15
[no jira]: Updating BRS information

olliecurtis:


docs/src/pages/BackpackReactScriptsPage/BackpackReactScriptsPage.js
```diff
@@ -109,16 +109,13 @@ npm start`}
       <Paragraph>
         <Code>backpack-react-scripts</Code> is aimed at providing teams and
         codebases with a standardised way of building production web apps with
-        Backpack + React at Skyscanner. When you create an app using{' '}
-        <Code>
-          npx create-react-app@2.1.2 my-app
```
olliecurtis, line 114: Removed this as I noticed that as this is in a `Code` block it does not wrap and the window does not scroll for you to be able to see the full command.

Plus referencing the top of the page command means that there is only one instance of this command so means it only needs to be managed in one place 😄

PR #645, opened 2020-08-24
[BPK-4141] Split DocsPageBuilder

shaundon:
There are no screenshots on purpose, so I know you've tested it locally 😉 

## What I've done

I started out by trying to simplify `DocsPageBuilder`, but it was so complex I was losing my mind looking into the abyss. So, I created a new component instead named `ComponentPage`. Instead of refactoring `DocsPageBuilder`, it's more like I'm creating a series of simpler components that will eventually deprecate it.

`ComponentPage` is a simple component. All it does is generate the table of contents, apply some outer styles to everything, renders the readme in the right place, and then displays any examples. It also has an `additionalContent` prop for adding arbitrary content below the readme (for example, a link to a Sassdoc).

The intention is to make `ComponentPage` as dumb as possible. It shouldn't need to know about screenshots, or anything that's specific to any particular implementation. All it does it take examples, a readme and optional additional content and then renders them nicely.

Then, we have components that wrap `ComponentPage`:

### `NativeComponentPage`
Pretty much all it does is turn the `screenshots` prop into some rendered examples to pass to `ComponentPage`.

### `AndroidComponentPage`
Basically the same as `NativeComponentPage`, but it passes in the link to the Android docs as the `additionalContent` prop.

### `IOSComponentPage`
Basically the same as `NativeComponentPage`, but it passes in the link to the iOS docs as the `additionalContent` prop.

### `WebComponentPage`

Similar to the others, but for live examples as opposed to screenshots.

---

From reading this, you'll notice there's some duplication between the components, and might wonder if we should merge some of them. I don't think we should. Because the components are so small, the duplication is very minimal and I think keeping them separate helps keep a tidy mental model that these are all separate concerns.

There are two ways I've been able to keep the components very simple:

1. A lot of the props in `DocsPageBuilder` are optional, because it has to support so many use cases. E.g. `tokenTable`, `showMenu`, etc. Because `ComponentPage` only cares about showing component pages, we can make assumptions, for example there will always be a readme, and we always want the table of contents to display. This makes for a much simpler component.
1. `DocsPageBuilder` and the things surrounding it are way too complex. For example there's a whole component just for headings. I'm relying on standard Backpack components as much as possible so we don't need to define more things.

## Other things

* As discussed, I improved the table of contents.

## Future optimisations / out of scope for now

There is some extra refactoring I _could_ have done but elected not to, to avoid turning this into a bigger task than is necessary:

* We shouldn't need to pass in the device name, alt text or dimensions for screenshots. They never change so it should be centralised somewhere. **However**, if we change this it will mean updating every mobile component page we have, so the cost vs benefit scale would be off.
* Token pages use `DocsPageBuilder`. I've decided to leave those as is for now as it'll be too complex to refactor, and it's out of scope for what we're trying to achieve with this work.
* The design pages. There's not that much point as right now every single one of them is just a placeholder.
* I noticed all the licence headers say '2019' on them. Those can be done in a future PR.


docs/src/components/ComponentPage/AndroidComponentPage.js
```diff
@@ -0,0 +1,78 @@
...
+    <ComponentPage
+      usageTable={usageTable}
+      examples={screenshotsAsExamples}
+      readme={readme}
+      additionalContent={[
+        {
+          id: 'docs',
+          title: 'Class reference',
+          content: (
+            <BpkLink href={`/android/versions/latest/${documentationId}`} blank>
```
olliecurtis, line 68: Are we concerned that if the documentationId does not exist for some reason it would link to an empty page or not found?

georgegillams:
> I don't think we should. Because the components are so small, the duplication is very minimal and I think keeping them separate helps keep a tidy mental model that these are all separate concerns.

Also because premature optimisation is the root of all evil in the world?

docs/src/components/ComponentPage/AndroidComponentPage.js
```diff
@@ -0,0 +1,78 @@
...
+
+import ComponentPage from './ComponentPage';
+
+type Props = {
+  documentationId: string,
+  readme: string,
+  screenshots: [
+    {
+      id: string,
+      screenshots: {
```
k0nserv, line 34: Why is this `screenshots`(plural) when the value is an object(singular)

docs/src/components/ComponentPage/ComponentPage.js
```diff
@@ -0,0 +1,186 @@
...
+    dos: Array<string>,
+    donts: Array<string>,
+  },
+};
+
+const ComponentPage = (props: Props) => {
+  const { additionalContent, examples, readme, usageTable } = props;
+  return (
+    <div>
+      {/* Table of contents */}
```
olliecurtis, line 61: This has probably been done with reason but wondering instead of just having all of these in one page with comments to explain what they do, could they be broken down into components that are then included on the page just to help with future maintenance and make the file lighter?

Just an idea 😄

docs/src/components/ComponentPage/IOSComponentPage.js
```diff
@@ -0,0 +1,81 @@
...
+      examples={screenshotsAsExamples}
+      readme={readme}
+      usageTable={usageTable}
+      additionalContent={[
+        {
+          id: 'docs',
+          title: 'Class reference',
+          content: (
+            <BpkLink
+              href={`/ios/versions/latest/Classes/${documentationId}.html`}
```
olliecurtis, line 69: Are we concerned that if the documentationId does not exist for some reason it would link to an empty page or not found?

docs/src/components/ComponentPage/WebComponentPage.js
```diff
@@ -0,0 +1,81 @@
...
+    <ComponentPage
+      usageTable={usageTable}
+      examples={mappedExamples}
+      readme={readme}
+      additionalContent={[
+        {
+          id: 'docs',
+          title: 'Sassdoc',
+          content: (
+            <BpkLink href={`/sassdoc/#${sassdocId}`} blank>
```
olliecurtis, line 71: Are we concerned that if the documentationId does not exist for some reason it would link to an empty page or not found?

docs/src/components/ComponentPage/WebComponentPage.js
```diff
@@ -0,0 +1,81 @@
...
+type Props = {
+  examples: Array<{
+    id: string,
+    dark?: boolean,
+    title: string,
+    blurb: Node,
+    examples: Node,
+  }>,
+  readme: string,
+  sassdocId: string,
```
olliecurtis, line 37: Given that not every component would have a sassdocId should this be: `?string` as `null` might be passed through?

docs/src/components/ComponentPage/ComponentPage.js
```diff
@@ -0,0 +1,186 @@
...
+    blurb?: ContentType,
+    content?: ContentType,
+  }>,
+  readme: string,
+  additionalContent: ?Array<{
+    id: string,
+    title: string,
+    content: ContentType,
+  }>,
+  usageTable: ?{
```
k0nserv, line 51: Doesnt' this require us to explicitly pass `null` and thus isn't the type

```js
usageTable?: { 
  dos: Array<string>,
  donts: Array<string>.
}
```

better?

shaundon:
> Also because premature optimisation is the root of all evil in the world?

I heard a rumour that premature optimisation shot JFK

olliecurtis (changes requested):
A few comments with maybe some changes, will test locally to try out.

docs/src/components/ComponentPage/WebComponentPage.js
shaundon: Good catch, will change

shaundon: done

docs/src/components/ComponentPage/AndroidComponentPage.js
```diff
@@ -0,0 +1,78 @@
...
+    },
+  ],
+  usageTable: ?{
+    dos: Array<string>,
+    donts: Array<string>,
+  },
+};
+
+const AndroidComponentPage = (props: Props) => {
+  const { documentationId, screenshots, readme, usageTable } = props;
```
georgegillams, line 54: Can we spread `...rest` on to the outer ComponentPage instead of repeating things?

docs/src/components/ComponentPage/WebComponentPage.js
shaundon: Good catch, I'm adding a null check to prevent this.

docs/src/components/ComponentPage/AndroidComponentPage.js
```diff
@@ -0,0 +1,78 @@
...
+        {
+          id: 'docs',
+          title: 'Class reference',
+          content: (
+            <BpkLink href={`/android/versions/latest/${documentationId}`} blank>
+              View on Backpack&apos;s Android documentation
+            </BpkLink>
+          ),
+        },
+      ]}
```
georgegillams, line 75: If additional content was added by the consumer, shouldn't it be added here too?

docs/src/components/ComponentPage/ComponentPage.js
georgegillams: Are you thinking something like this?
```
<TableOfContents name="Examples" items={examples} />
<TableOfContents name="Additional content" items={additionalContent} />
```

shaundon: I considered it, but I think it would actually **increase** maintenance. Right now it's a longish file, but the contents are very simple. You've got:

1. A few instances of `BpkList` to generate the table of contents
1. A few blocks / `array.map`s to generate the content

If there were complex logic involved in those things, then I think there'd be a better argument for splitting it, but it's really just null checks and iterators.

 Splitting brings its own maintenance burden (it would need props defining and so on) and I don't think the benefits are greater than the cost.

docs/src/components/ComponentPage/IOSComponentPage.js
georgegillams: Flow should enforce this though, right?

docs/src/components/ComponentPage/AndroidComponentPage.js
shaundon: Right now I'm operating under the assumption that every Android page has an accompanying docs page link.

I think it's preferable to keep using that assumption until it's proven wrong (we'd discover the error very quickly when we try to roll this out to all pages) than to add a null check that might not be needed.

docs/src/components/ComponentPage/ComponentPage.js
olliecurtis: No problem, with that reasoning I am happy to have this the way it is, just wanted to raise this incase it had been missed

docs/src/components/ComponentPage/IOSComponentPage.js
shaundon: Right now I'm operating under the assumption that every iOS page has an accompanying docs page link.

I think it's preferable to keep using that assumption until it's proven wrong (we'd discover the error very quickly when we try to roll this out to all pages) than to add a null check that might not be needed.

docs/src/components/ComponentPage/ComponentPage.js
shaundon: Yeah, always good to check these assumptions!

docs/src/components/ComponentPage/AndroidComponentPage.js
olliecurtis: I guess but its better to be preventative and it not be an issue than, let it roll out and give our consumers a bad experience, then have to do a fix rather than have something in there that would prevent issues and broken functionality in the first place.

Plus we are already doing it for the Web page so it would kind of make sense to do it here?

docs/src/components/ComponentPage/IOSComponentPage.js
olliecurtis: I guess but its better to be preventative and it not be an issue than, let it roll out and give our consumers a bad experience, then have to do a fix rather than have something in there that would prevent issues and broken functionality in the first place.

Plus we are already doing it for the Web page so it would kind of make sense to do it here?

shaundon: There's no chance of this affecting consumers, because if this does turn out to be a problem it'll immediately be caught during development when we try to use these new components on the remaining pages.

The web is a different case, because I have the opposite assumption there — that not all components have an accompanying sassdoc.

shaundon: Philosophically, I'd prefer to put something in the code that will cause a crash _during development_ so we can find and fix it, than put in a null check that'll fail quietly, which we might not notice.

This article (which is about Swift but the logic applies here) goes into more detail: https://wolfmcnally.com/82/force-unwrapping-swift-not-bad-thing

docs/src/components/ComponentPage/ComponentPage.js
shaundon: Yeah you're right, will change.

shaundon: Ah I just remembered why I did it this way 😓 

Screen Shot 2020-08-25 at 17 29 18

Obviously I can add the `defaultProps` declaration, but what's the point if the former way works?

docs/src/components/ComponentPage/AndroidComponentPage.js
shaundon: The Flow type was wrong here, I should have made it an array. I expect it wasn't caught because the file that consumes this doesn't have the Flow annotation. I'm fixing that now (and for the other files I created/modified)

docs/src/components/ComponentPage/AndroidComponentPage.js
shaundon: Gonna resolve this conversation, as the same discussion applies to both Android and iOS, so we can just centralise it on the iOS thread

docs/src/components/ComponentPage/WebComponentPage.js
shaundon: done

docs/src/components/ComponentPage/AndroidComponentPage.js
shaundon: done

georgegillams (commented):
I love it!
Could of things.

I think this stood out better when we included the logo:
Screenshot 2020-08-25 at 17 39 11

Also `Implementation` seems like a strange heading. Could we make the readme be `Usage` or even just `Readme`?

shaundon:
@georgegillams 

> Also Implementation seems like a strange heading. Could we make the readme be Usage or even just Readme?

This was discussed in the thread we had on Slack about it. I was originally going to put 'Readme', but then @olliecurtis suggested 'Usage' or 'Implementation'. I went with 'Usage', but the readmes already have a heading called that so it was repetitive. So I settled on 'Implementation'. Happy to take more suggestions though, I don't really mind what we call it.

shaundon:
> I think this stood out better when we included the logo:

You're probably right, but it was just much simpler to not include it. We don't have it for the other documentation things so it was just easier to leave it out. Does it actually **need** to stand out? Including it in the table of contents will do 100x more for discoverability than the logo 😉

docs/src/components/ComponentPage/AndroidComponentPage.js
shaundon: I originally did do this, but then I was concerned it would violate principle of least surprise by not making it obvious that `readme` and `usageTable` were being passed through, you'd only know they were there from looking at the Flow type. 

If you feel strongly about it I don't mind changing it, but that was my rationale.

georgegillams: I see it like `AndroidComponentPage` is effectively a subclass of `ComponentPage`, so it could be inferred, however explicit won't do any harm either.

docs/src/components/ComponentPage/AndroidComponentPage.js
shaundon: There is no additional content a consumer can add to this component. 

When we add the ability for pages to display more things (e.g. a link to view the repo on Jetpack) we'll add it as another prop which is then passed as part of the `additionalContent` array.

docs/src/components/ComponentPage/IOSComponentPage.js
shaundon: @georgegillams yeah Flow will enforce it, which is why it can't affect any consumers because it'll be caught during development.

docs/src/components/ComponentPage/AndroidComponentPage.js
georgegillams: Oh so `additionalContent` is strictly for the 4 page types to use? That seems overly restrictive. Would it make more sense to adopt a compositional approach, where the Pages accept an array of anything, and we create common elements to pass in?

georgegillams: For example, `usageTable` isn't specific to Android pages, so why don't we just pass in a UsageTable node to `additionalContent`?

shaundon: > Oh so additionalContent is strictly for the 4 page types to use?
I don't really understand the question, because every prop on `ComponentPage` is for these pages to use.

> For example, usageTable isn't specific to Android pages, so why don't we just pass in a UsageTable node to additionalContent?

I did want to do this, where `ComponentPage` would just take `content` and then would only render it, as well as generate a table of contents. It's what I started out doing initially.

I changed to the route I'm proposing now because `usageTable` needs to be at the top of the page, and passing it into `additionalContent` would put it at the bottom. This is also why I asked on Slack about moving the usage table to the bottom, to enable this API.

shaundon: You could, however, make this argument that `readme` should be passed in as `additionalContent`. The downside would be that the logic to convert it to markdown would be duplicated in four places. What do you reckon?

georgegillams: Ah right yeah that makes sense - but what about other sections in the future. Like if we want a `copy` section like Luna has, we'd have to add that to all 4 page types too?

georgegillams: Yeah for README I guess that makes sense too. The duplicated logic could just be a method call, so seems fine to me

shaundon: I've not heard of Luna but I googled it and assume it's [this one](https://luna.sainsburys.co.uk/components/autocomplete), let me know if I've got the wrong one.

Yeah so in this scenario there's two ways to do it:

**First**
Add a `copy` prop to `ComponentPage`, which the four page types will pass down (like how we have done readme).

**Second**
Add a `copy` prop to each of the platform-specific component pages, and then pass it down as part of the additional content array.

I'd advocate for the first option. I feel like if it's a prop that can apply to all four platforms it should be centralised (like how readme is now), and when it's something specific to a particular platform it should be passed in as additional content (like how the links to docs are).

Here's a more true to life example based on our roadmap. We want to add links to view iOS components on Cocoapods, and we also want to add a link to view the source code of components on GitHub. 

Because the Cocoapods one is specific to iOS I would pass it as part of `additionalContent`, but for GitHub we know it applies to all platforms, so I'd add it as a prop to `ComponentPage`.

georgegillams: For either of those examples, you could leave the Pages unmodified and consumers could use:

```
<AndroidPage ... additionalContent={[
  githubLink("/src/..."),
} />
```

That way we are composing the page from reusable sections, and whatever sections we wish to add in the future we can without turning this class into a monster

shaundon: Sorry for replying to this at such an unsociable time, but I've just come on my laptop to do something else and saw the comment (no need to reply until tomorrow).

I think you're right, but I also think maybe you've missed one aspect. Or I have 😅 I'll DM you tomorrow morning and we can talk it through on Zoom, will be easier than going back and forth on here

georgegillams: Sounds great 👍

georgegillams (approved):

docs/src/components/ComponentPage/ComponentPage.js
k0nserv: My understanding was that this is how you are recommended to do it. But here's what the docs [says](https://flow.org/en/docs/react/components/#toc-using-default-props)

shaundon: Ah interesting. Tbh I don't have a strong opinion on this. My philosophy with Flow is "whatever it takes to get it to stop complaining". So I can either change it and add `defaultProps`, or leave it as is. 

I lean towards leaving it, because it's less code, but as I say, I don't mind. What do you think?

docs/src/components/ComponentPage/IOSComponentPage.js
olliecurtis: Flow would capture it if the prop wasn't but what happens if we assume there is a doc id for a component but once its deployed, the link is generated based on the assumption but actually doesn't exist in the iOS/Android docs for whatever reason?

Flow can't actually test the url is valid and works but its just a small consideration to take into account here

shaundon: This is true, but I don't think that's relevant to this discussion as having a null check or not won't affect that.

olliecurtis (approved):

PR #649, opened 2020-08-27
[BPK-4141] Finish moving components to new page builder

shaundon:
The only uses of `DocsPageBuilder` left are in non-component pages, which are out of scope for this task:

Screen Shot 2020-08-27 at 13 50 07

At some point in the future we can create a new page builder component to handle these non-component use cases, and then kill `DocsPageBuilder`.

### DIFF
--- a/docs/src/components/ComponentPage/WebComponentPage.js
+++ b/docs/src/components/ComponentPage/WebComponentPage.js
@@ -34,7 +34,7 @@ type Props = {
     dark?: boolean,
     title: string,
     blurb: Node,
-    examples: Node,
+    examples: ?Node,
   }>,
   readme: ?string,
   sassdocId: ?string,
@@ -51,14 +51,14 @@ const WebComponentPage = (props: Props) => {
       id: example.id,
       title: example.title,
       blurb: example.blurb,
-      content: (
+      content: example.examples ? (
         <PresentationBlock
           whiteBackground={!example.dark}
           darkBackground={example.dark}
         >
           {example.examples}
         </PresentationBlock>
-      ),
+      ) : null,
     };
   });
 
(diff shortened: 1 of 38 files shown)


docs/src/components/ComponentPage/WebComponentPage.js
```diff
@@ -51,14 +51,14 @@ const WebComponentPage = (props: Props) => {
       id: example.id,
       title: example.title,
       blurb: example.blurb,
-      content: (
+      content: example.examples ? (
```
shaundon, line 54: Needed this because some examples will only have a blurb but no actual code — for example web images page.

georgegillams: I'm confused by this. [All the examples](https://backpack.github.io/components/image/?platform=web) have a rendered component

shaundon: Argh, yeah sorry. I meant the icon page! Form elements has this too.

docs/src/pages/WebSelectPage/WebSelectPage.js
```diff
@@ -175,22 +174,11 @@ const components = [
   },
 ];
 
-const blurb = [
-  <IntroBlurb>
-    Backpack selects override the default styles in most modern browsers. In
-    some older browsers they simply fall back to the browser default.
```
georgegillams, line 181: Where did this content go?

shaundon: I killed it —I was going to move it but honestly it felt redundant, the browser defaults thing is no longer a concern (we don't support any browsers where this fallback happens) and _selects override the default styles in most modern browsers._ felt like a redundant statement, that's just describing most of our components 😛 

Also, if it was that important, we don't mention it on other pages where it could be applicable (like text input)

georgegillams: Makes sense 👍

georgegillams (approved):

PR #650, opened 2020-08-28
[NO-JIRA] Fix Fuse usage in IE and Edge

georgegillams:
image


shaundon (approved):

shaundon:
I didn't run it locally, but as this breaks functionality in IE/Edge I'm happy to trust that you tested it!

PR #652, opened 2020-08-28
[BPK-4129] Add social meta

georgegillams:


georgegillams:
> Is this WIP or ready for review? If WIP, please label it

Just needs design review

shaundon:
I mean this in the nicest way possible, but you've just repeated what the label said 😂 

Are you actively in the process of getting a design review? If not, I would argue that it's still WIP as it's not mergeable

georgegillams:
> I mean this in the nicest way possible, but you've just repeated what the label said 😂
> 
> Are you actively in the process of getting a design review? If not, I would argue that it's still WIP as it's not mergeable

tbh I think it can just be reviewed, and we can iterate on the design later if we see fit. Mainly saying that as it's hard to see how these values will appear on social platforms until they're live

shaundon (approved):

shaundon:
Yeah I agree. I think it's also fine for us to do the 'ask forgiveness over permission' thing with design decisions like this. We don't want to have to ask Dima for everything, so as long as we make sensible choices that follow the brand guidelines (as you have here) we can default to assuming it's ok.

PR #704, opened 2020-09-16
[No Jira] Make Dependabot only handle direct dependencies

shaundon:
Looked into the Dependabot config and here's what I've found out:

- We can't currently make it ignore patches, like we wanted
- But we can at least make it ignore indirect dependencies (deps of deps)

I like the idea of making it ignore indirects, because that's how a human would do it — we only manage the deps explicitly defined in our package.json, we don't generally upgrade deps of deps (security fixes notwithstanding), so I made it do that.

As with all of these, we won't know the full effect until we try it out and see how it performs.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -9,5 +9,7 @@ updates:
   labels:
   - dependabot
   versioning-strategy: increase-if-necessary
+  allow:
+    - dependency-type: "direct"
   ignore:
     - dependency-name: "bpk-*"


.github/dependabot.yml
```diff
@@ -9,5 +9,7 @@ updates:
   labels:
   - dependabot
   versioning-strategy: increase-if-necessary
+  allow:
+    - dependency-type: "direct"
```
georgegillams, line 13: What about security patches in transient-dependencies?

shaundon: I think Snyk takes care of those, right?

olliecurtis: Yes that is correct!

olliecurtis: Plus when we run an npm install in this repo anyway we update the package-lock file so any patches get picked up with this

olliecurtis (approved):

PR #708, opened 2020-09-16
[No Jira] Update eslint-config-skyscanner

shaundon:
Methodology:

- Fix most errors
- Ignore all the image ones, there are too many of them and it's not worth the time investment to fix them. I did try updating some to use `BpkImage` but they broke.

### DIFF
--- a/docs/src/components/ApidocLink/ApidocLink.js
+++ b/docs/src/components/ApidocLink/ApidocLink.js
@@ -105,11 +105,13 @@ export const createFromType = (type, defProps) => {
 const ApidocLink = ({ logo, renderLink }) => (
   <aside className={getClassName('bpkdocs-apidoc-link')}>
     {logo && (
+      /* eslint-disable backpack/use-components */
       <img
         className={getClassName('bpkdocs-apidoc-link__logo')}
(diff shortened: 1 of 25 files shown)


docs/src/components/ApidocLink/ApidocLink.js
```diff
@@ -105,11 +105,13 @@ export const createFromType = (type, defProps) => {
 const ApidocLink = ({ logo, renderLink }) => (
   <aside className={getClassName('bpkdocs-apidoc-link')}>
     {logo && (
+      /* eslint-disable backpack/use-components */
```
georgegillams, line 108: Does this work? Might be neater?

```suggestion
      // eslint-disable-next-line backpack/use-components
```

olliecurtis (approved):

docs/src/components/ApidocLink/ApidocLink.js
shaundon: Ah yeah, probably. The thing is this is total laziness of me but I don't want to have to update all 20ish files to this format 🙈

georgegillams (approved):

PR #726, opened 2020-10-01
[NO-JIRA] Fix 0-length example arrays

georgegillams:


olliecurtis (commented):
LGTM - I am just curious though why/when would we not have examples if we are making a page?

As we would define a title and blurb for the page but not provide examples for that, just seems a bit strange. Wdyt?

georgegillams:
> LGTM - I am just curious though why/when would we not have examples if we are making a page?
> 
> As we would define a title and blurb for the page but not provide examples for that, just seems a bit strange. Wdyt?

Remember this?
IMG_6238

It's because we have `examples: []`

olliecurtis:
Yeah I remember this issue - I guess the question here is why do we have this without an example? If this is just text shouldn't it live in the blurb or in the README instead?

olliecurtis (approved):

georgegillams:
This is the whole code:
```
 {
    id: 'icon-alignment',
    title: 'Icon alignment within buttons',
    blurb: [
      <Paragraph>
        The <BpkLink href={ROUTES.ALIGNMENT}>Alignment</BpkLink> page gives
        examples of icon alignment using HOCs provided in the{' '}
        <Code>bpk-component-icon</Code> package.
      </Paragraph>,
    ],
    examples: [],
  },
```

So we could fix it by removing `examples: []`, but it's possible we'll do this in the future so imo it's best to handle it properly

PR #749, opened 2020-10-13
[BPK-4227] Add web overlay docs

shaundon:
Screen Shot 2020-10-13 at 17 23 42

### DIFF
--- a/docs/src/layouts/links.js
+++ b/docs/src/layouts/links.js
@@ -159,7 +159,8 @@ const ComponentsLinks = [
         id: 'OVERLAY',
         route: routes.OVERLAY,
         children: 'Overlay',
-        tags: ['android'],
+        tags: ['android', 'web'],
+        keywords: ['tint', 'scrim'],
       },
       {
         id: 'CARD',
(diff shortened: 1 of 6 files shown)


olliecurtis (approved):

georgegillams:
Screenshot 2020-10-13 at 17 31 40

This got me thinking - should we apply the radius and `overflow: hidden` to the wrapper, so that consumers don't need to worry about getting the background corner-radius to match exactly?

georgegillams:
> Ah yeah this is quite an interesting one! I did initially do that, using `border-radius: inherit`, and it worked in situations where the element the overlay is contained in has a border radius. However it didn't work in instances where the content **inside** the overlay was edge-to-edge and had rounded corners, for example the images you see on this docs page. That's because `border-radius: inherit` only goes down the cascade, not up. So long story short I made it an explicit prop.

I've just tested it out, and the consumer could actually use a className of their own to apply a `border-radius` and `overflow` style. So I don't think we need to change the component as they have both options.

Maybe we should make a note in the docs that this is also an option? Or just leave it as is

shaundon:
I thought about this too, but it felt like it would be difficult to document well, as there's some nuance between the border-radius coming from the parent or the child, so it felt like a better consumer experience to just do it explicitly

PR #750, opened 2020-10-19
[BPK-4242] Docs for updated RN chip component

shaundon:
I also updated the screenshots CSS to make it look good when more than two screenshots are displayed for an example. You can see the result of it in the screenshot.

Screen Shot 2020-10-19 at 10 22 14

### DIFF
--- a/docs/src/components/DocsPageBuilder/ComponentScreenshots.scss
+++ b/docs/src/components/DocsPageBuilder/ComponentScreenshots.scss
@@ -23,6 +23,7 @@
 
   @include bpk-breakpoint-above-mobile {
     display: flex;
+    flex-wrap: wrap;
   }
 
   &__item {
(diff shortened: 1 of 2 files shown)


docs/src/components/DocsPageBuilder/ComponentScreenshots.scss
```diff
@@ -31,6 +32,7 @@
 
     @include bpk-breakpoint-above-mobile {
       width: calc(50% - #{$bpk-spacing-base});
+      min-width: calc(50% - #{$bpk-spacing-base});
```
georgegillams, line 35: So this makes them larger if the viewport is wide enough, right?

shaundon: Yeah they should be exactly the same as before, but now when you have more than two it won't make them tiny, it'll wrap them onto a new row

georgegillams (approved):

PR #765, opened 2020-10-26
[BPK-4252] Add MarkdownPage and migrate status page to use it

shaundon:
Proof of concept of `MarkdownPage`, with the status page migrated to use it.

## Answers to questions you might have 

**Why are `title` and `subtitle` props instead of just coming from the markdown?**
We need to know the title so we can set `window.title`. I tried pulling it out automatically but it's hacky so I decided to be explicit instead. This way also allows us to put that nice white area at the top with the title and subtitle. When we expand this PoC to do some snazzy automatic generation stuff, we should be able to pull these out of front matter or something to set them, so it should scale.

**The background alternating colours don't work with these**
Nope. YAGNI.

Screen Shot 2020-10-26 at 16 41 43

### DIFF
--- a/docs/src/components/MarkdownPage/MarkdownPage.js
+++ b/docs/src/components/MarkdownPage/MarkdownPage.js
@@ -1,0 +1,60 @@
+/*
+ * Backpack - Skyscanner's Design System
+ *
+ * Copyright 2016-2019 Skyscanner Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/* @flow strict */
+
+import React from 'react';
+import { cssModules } from 'bpk-react-utils';
+import BpkContentContainer from 'bpk-component-content-container';
+import Helmet from 'react-helmet';
+
+import BpkMarkdownRenderer from '../DocsPageBuilder/BpkMarkdownRenderer';
+import Heading from '../Heading';
+import IntroBlurb from '../IntroBlurb';
+
+import STYLES from './MarkdownPage.scss';
+
+const getClassName = cssModules(STYLES);
+
+type Props = {
+  title: string,
+  subtitle: ?string,
+  content: string,
+};
+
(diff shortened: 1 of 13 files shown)


docs/src/components/MarkdownPage/MarkdownPage.js
```diff
@@ -0,0 +1,60 @@
...
+import Heading from '../Heading';
+import IntroBlurb from '../IntroBlurb';
+
+import STYLES from './MarkdownPage.scss';
+
+const getClassName = cssModules(STYLES);
+
+type Props = {
+  title: string,
+  subtitle: ?string,
```
georgegillams, line 36: I don't think this should be called `subtitle`, because it's a paragraph

shaundon: There's nothing that says subtitles are a single line. This usage meets the dictionary definition almost perfectly:

Screen Shot 2020-10-26 at 16 59 32

georgegillams: Actually that makes sense 😂

georgegillams:
The headings look a bit big. They're as big as the main one at the top 🤔 
Are we planing to lose the alternating section colouring? Seems it would make every page look quite bland

docs/src/components/MarkdownPage/MarkdownPage.js
shaundon: You know me and grammar 😛

shaundon:
> The headings look a bit big. They're as big as the main one at the top 🤔

They're the same size as on non-markdown pages

> Are we planing to lose the alternating section colouring? Seems it would make every page look quite bland

Yep, this is addressed in the PR description. The alternating sections are nice but they'd be a nightmare to implement when the page is being automatically generated from markdown, so I'd rather lose them than hack them back in.

shaundon:
> Are we planing to lose the alternating section colouring? Seems it would make every page look quite bland

Another way I think of this is that this is first-and-foremost a documentation site, so the content is king. We should make it look good but it's a secondary concern compared to ease of use (both for users and us maintaining it).

Tbh I want to rip a few things out of it, those massive images that push all the content below the fold are ridiculous 🙈

georgegillams:
Oh yeah they are. Everything just looks a bit tight and garish though 🤔

shaundon:
> Everything just looks a bit tight and garish though

In the nicest way possible, I just don't think it matters. Not for a PoC anyway. We can look at increasing the margin around headings later if it's really a problem.

Also bear in mind this is a strange-looking page compared to most others where the headings will be more spaced out with content.

georgegillams:
As for the massive images, as far as I'm concerned we can kill them 🎉

shaundon:
@georgegillams I feel like I've addressed your questions — is this ok to approve or do you still have concerns?

georgegillams (approved):

PR #768, opened 2020-10-29
[NO-JIRA] Markdown alternating sections

georgegillams:
Did this mostly for fun, and not for profit 😂

I think we decided this was yagni, but I wanted the challenge 🤪

It should only be needed in this one place, so I think the work to maintain this going forward will be minimal.

Screenshot_2020-10-28 Backpack Status Backpack(1)

### DIFF
--- a/docs/src/components/MarkdownPage/MarkdownPage.js
+++ b/docs/src/components/MarkdownPage/MarkdownPage.js
@@ -23,10 +23,12 @@ import { cssModules } from 'bpk-react-utils';
 import BpkContentContainer from 'bpk-component-content-container';
 import Helmet from 'react-helmet';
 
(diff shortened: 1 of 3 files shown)


shaundon:
thanks i hate it

shaundon:
This does look good but my rationale from the other day hasn't changed:

I said we probably could reinstate the alternating sections but it would be hacky and increase maintenance burden. This is hacky and increases the maintenance burden.

Also, it's a documentation website, it doesn't need to be this stylised. It *should* have a simple design.

olliecurtis (commented):
LGTM - Just gonna run it locally

georgegillams:
> LGTM - Just gonna run it locally

See @shaundon's comment above ;)

georgegillams:
Killing this as it's YAGNI

PR #776, opened 2020-10-30
[BPK-4271] Add a static site generator

shaundon:
As part of `npm run docs`, any markdown files placed in `docs/src/static-pages` will be transformed into docs site pages and will show up in the sidebar and so on.

Right now it only works for `using`-type pages but we can expand it over time with a trivial amount of effort. This will not only enable more easy contributions from others, it'll also make it much easier for us to update the site's content.

As part of this I migrated the 'Status' page to this new format, you can see below it's all good!

Note that this is intended to be the first iteration of this. It works great but we may tweak and add more functionality over time.

Screen Shot 2020-10-30 at 18 00 20


.eslintrc
```diff
@@ -23,6 +23,7 @@
       }
     }
   },
+  "ignorePatterns": ["**/generated/**/*.js", "scripts/build-process/templates/**/*.js"],
```
shaundon, line 26: We don't want to lint automatically generated stuff, and the template files.

docs/src/constants/routes.js
```diff
@@ -28,7 +28,6 @@ export const GETTING_STARTED = '/using/getting-started';
 export const BACKPACK_REACT_SCRIPTS = '/using/backpack-react-scripts';
 export const BASE_STYLESHEET = '/using/base-stylesheet';
 export const PRINCIPLES = '/using/principles';
-export const STATUS = '/using/status';
```
shaundon, line 31: No longer needed because the autogeneration handles it.

docs/src/layouts/links.js
```diff
@@ -499,11 +501,7 @@ export default [
         route: routes.BASE_STYLESHEET,
         children: 'Base stylesheet',
       },
-      {
-        id: 'STATUS',
-        route: routes.STATUS,
-        children: 'Backpack Status',
-      },
+      ...autogeneratedLinks.using,
```
shaundon, line 504: In the future we'd add more instances of this, e.g. a `.guidelines` one.

package.json
```diff
@@ -73,6 +74,7 @@
     "@babel/preset-flow": "^7.10.4",
     "@babel/preset-react": "^7.10.4",
     "@babel/register": "^7.10.4",
+    "@github-docs/frontmatter": "^1.3.1",
```
shaundon, line 77: Used for parsing the frontmatter info from pages.

georgegillams (approved):

PR #781, opened 2020-11-02
[BPK-4274] Replace all the 'using backpack' pages with their non-union markdown equivalents

shaundon:
Will be invisible to end users, so I didn't include a screenshot as it would just look the same as the current site


docs/src/constants/redirect-routes.js
```diff
@@ -135,7 +135,6 @@ export default {
   '/components/web/tickets': ROUTES.TICKET,
   '/components/web/tooltips': ROUTES.TOOLTIP,
   '/components/web/typography': ROUTES.TEXT,
-  '/resources': ROUTES.GETTING_STARTED,
```
shaundon, line 138: Was simpler to remove this redirect than figure out how to dynamically populate it. The redirects are really niche anyway so I doubt this ever gets hit.

docs/src/pages/HomePage/HomePage.js
```diff
@@ -75,7 +75,7 @@ const CARD_CONTENTS = [
   {
     key: 'using-backpack',
     title: 'Using Backpack',
-    href: ROUTES.GETTING_STARTED,
+    href: ROUTES.USING_BACKPACK,
```
shaundon, line 78: Pointing this to the root (`/using`) rather than (`/using/getting-started`) because the latter now comes from autogen and we won't know it ahead of time.

docs/src/routes/Routes.js
```diff
@@ -132,20 +128,8 @@ export const ROUTES_MAPPINGS = [
...
-      },
-      {
-        path: ROUTES.BACKPACK_REACT_SCRIPTS,
-        component: BackpackReactScriptsPage,
-      },
-      { path: ROUTES.BASE_STYLESHEET, component: BaseStylesheetPage },
-      { path: ROUTES.PRINCIPLES, component: PrinciplesPage },
-      ...autogeneratedUsingPages,
-    ],
+    redirect: autogeneratedUsingPages[0].path,
```
shaundon, line 131: I figured this was an ok way to make redirects work without having to hard code the path of an autogen file in here. It'll just redirect users to whatever the first autogen page is.

shaundon:
Screen Shot 2020-11-02 at 12 24 44

Look how much code we can delete! 🎉

docs/src/constants/redirect-routes.js
georgegillams: These are essentially just strings though 🤔 
So you could do `'/resources': '/using',`

shaundon: Yeah I suppose we could just to `ROUTES.using` but I would wager that nobody has hit this route in 2020 so I don't think the redirect is still necessary

georgegillams (approved):

docs/src/constants/redirect-routes.js
georgegillams: Yeah that's fair

shaundon: We can call it 'ruthless prioritisation' 😉

PR #799, opened 2020-11-09
[BPK-4275] Add edit links to auto generated pages

shaundon:
Will make writing the guide easier. I know it's not perfect atm, e.g. it could use a link to the new contributing guidelines but we can add that once we have them.

Screen Shot 2020-11-09 at 10 54 42

### DIFF
--- a/scripts/build-process/generate-pages-from-markdown.js
+++ b/scripts/build-process/generate-pages-from-markdown.js
@@ -23,9 +23,9 @@ const fs = require('fs');
 const colors = require('colors');
 const frontmatter = require('@github-docs/frontmatter');
 
-const PATH_TO_MARKDOWN_FILES = './docs/src/static-pages';
-const PATH_TO_ROUTES_FILE = './docs/src/constants/generated/routes.js';
-const PATH_TO_LINKS_FILE = './docs/src/layouts/generated/links.js';
+const PATH_TO_MARKDOWN_FILES = 'docs/src/static-pages';
+const PATH_TO_ROUTES_FILE = 'docs/src/constants/generated/routes.js';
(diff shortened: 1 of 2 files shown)


olliecurtis (commented):
LGTM! Just going to run it locally

scripts/build-process/generate-pages-from-markdown.js
```diff
@@ -106,13 +107,20 @@ const enrichedMarkdownFiles = markdownFiles.map(fileName => {
...
     // Escape backticks to prevent them terminating the string.
-    const escapedContent = content.replace(/`/g, '\\`');
+    let escapedContent = content.replace(/`/g, '\\`');
+
+    // Add a footer to the bottom of the content for editing.
+    escapedContent += `
+## Improve this page
+
+[Edit this page on GitHub](https://github.com/skyscanner/backpack-docs/edit/master/${PATH_TO_MARKDOWN_FILES}/${fileName})
+    `;
```
georgegillams, line 123: I feel like this should maybe be part of the MarkdownPageRenderer rather than just shoving on some extra MD. I guess both are valid though.

What happens if the MD file doesn't end with a line break?

olliecurtis (approved):

scripts/build-process/generate-pages-from-markdown.js
shaundon: Yeah we could potentially do that, I just felt like it was better in here because we have access to the filename more easily, and if a page is autogenerated we **know** it can be edited through this process, whereas the MarkdownPage component is agnostic, it just displays whatever its given.

> What happens if the MD file doesn't end with a line break?

I'll check and fix if any issues

shaundon: Seems to be fine

PR #812, opened 2020-11-13
[No Jira] Fix images that were too large

shaundon:
For now I've just changed the logic we're using in the CSS.

In the future we may want to add a plugin to the Markdown renderer so that we can specify extra props in our files. Jekyll supports this, so on my blog I can do this to add a custom class:

```
![Screenshots of Personal Best 2](/assets/post-images/personal-best-2-screenshots.png){:class="post-image"}
```

But this is good for now.


olliecurtis (commented):
LGTM - just wanna try it locally

olliecurtis (approved):
@shaundon has sent me screenshots separately so happy to approve as I have seen it working

PR #825, opened 2020-11-24
[BPK-4299] ARIA live docs

shaundon:
Would appreciate somebody running this locally to check the examples make sense.

May not be perfect but probably good enough for now? I'm wary of duplicating things on the accessibility guidelines.

Screen Shot 2020-11-24 at 15 20 30


docs/src/pages/components/AriaLive/AriaLivePage.js
```diff
@@ -0,0 +1,41 @@
...
+
+import Web from './WebAriaLivePage';
+
+const Page = () => (
+  <DocsPageWrapper
+    title="ARIA Live"
+    blurb={[
+      <IntroBlurb>
+        ARIA live regions are used to identify changes in a page&apos;s content
+        to assistive technologies.
```
georgegillams, line 34:
```suggestion
        ARIA live regions are used to notify assistive technologies of changing content.
```

shaundon: I like it!

docs/src/pages/components/AriaLive/WebAriaLivePage.js
```diff
@@ -0,0 +1,288 @@
...
+  }
+}
+
+const components = [
+  {
+    id: 'about',
+    title: 'About ARIA live regions',
+    blurb: [
+      <Paragraph>
+        Oftentimes, pages update without the content reloading completely. For
```
georgegillams, line 207: Is oftentimes a word?

docs/src/pages/components/AriaLive/WebAriaLivePage.js
```diff
@@ -0,0 +1,288 @@
...
+        Sometimes it&apos;s necessary for an ARIA live region to be visible to
+        everyone, not just assistive technologies. The
+        <BpkCode>visible</BpkCode> prop is available for these use cases.
+      </Paragraph>,
+      <Paragraph>
+        In this example, as the chips are toggled on and off, text describing
+        the changes appears in the ARIA live region below them. Ordinarily this
+        would be only visible to assistive technologies, but for demonstration
+        purposes the <BpkCode>visible</BpkCode>
+        prop has been supplied to the component to make it visible to all.
```
georgegillams, line 255: This looks like copy-pasta?

docs/src/pages/components/AriaLive/WebAriaLivePage.js
shaundon: Indubitably, but I'll change it for something more accessible

docs/src/pages/components/AriaLive/WebAriaLivePage.js
```diff
@@ -0,0 +1,281 @@
...
+              checked={direct}
+              onChange={this.toggleDirectness}
+            />
+          </div>
+        </div>
+        <BpkAriaLive
+          visible
+          id={this.id}
+          className={getClassName('bpkdocs-aria-live-page__aria-live')}
+        >
```
georgegillams, line 103: Because we are just changing the content that's there, a screen-reader will only read out the change. eg "Direct" or "Nara"
We should use `aria-atomic="true" to tell it we want to read the whole thing when a change is made.

We should also explain this in the example blurb

shaundon: Ahh, thanks! I knew I'd make some sort of mistake like that, I'll update it and the blurbs to reflect this.

georgegillams (approved):
Could you update the dependency to get the latest README? Then this is all good to merge 👍

PR #827, opened 2020-11-25
[BPK-4300]: Updating engineering A11y docs

olliecurtis:
Updating the docs with the latest version of the guidelines

### DIFF
--- a/docs/src/static-pages/accessibility.md
+++ b/docs/src/static-pages/accessibility.md
@@ -24,13 +24,72 @@ Modifies the `bpk-visually-hidden` mixin to allow the element to be focusable wh
 
 Hide visually and from screen readers, but maintain layout.
 
-# Color
+# Colour
+
+## Colour contrast
 
 ### **Why**
 
-If travellers have customised their device settings to suit their needs, we need to respect their preferences. 
+Text or UI elements that don't have high enough colour contrast with backgrounds can be very hard to read
 
 ### **Who**
+- Low vision
+- Colour blindness
+- Other vision impairments
+- Cognitive impairments that affect vision, like dyslexia
+- Extreme conditions, like bright sunlight
+
+### **How**
+
+Use colour pairings with high enough contrast to make text and important UI elements clearly visible against their background.
+
+- Small text (smaller than `18pt` or `14pt` bold) minimum ratio = 4.5:1
+- Large text (equal to or larger than `18pt` or `14pt` bold) minimum ratio = 3:1
+- UI components or meaningful graphics minimum ratio = 3.1
+
+### **Tips**
+
+Text is easier to read on solid colour backgrounds than backgrounds that are busy, gradated or an image
+
+### **Tests**
+
+- Use our [colour pairings chart]((/guidelines/colors/#section_pairings)) for accessible brand palette combinations
+- Use [Stark plugin](https://www.getstark.co/0) in Sketch or Figma to check colour contrast ratios
+- [Other contrast checkers](https://developer.paciellogroup.com/resources/contrastanalyser/) are good too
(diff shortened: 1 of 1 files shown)


docs/src/static-pages/accessibility.md
```diff
@@ -24,13 +24,72 @@ Modifies the `bpk-visually-hidden` mixin to allow the element to be focusable wh
...
+### **Tips**
+
+- Add text where possible to help explain meaning
+- Provide additional cues like colour keys or icons
+
+## Personalising colour
+
+### **Why**
+
+If travellers have customised their device settings to suit their needs, we need to respect their preferences
```
shaundon, line 89:
```suggestion
If travellers have customised their device settings to suit their needs, we need to respect their preferences.
```

docs/src/static-pages/accessibility.md
```diff
@@ -243,22 +280,47 @@ Using "headings" allows screen reader users to quickly understand the content of
...
+When screen readers come to a properly marked up list, they will notify the user that there is a list and how many items are in it.
+
+### **Who**
+
+- Screen reader users
+- Robots (eg search engines)
+
+### **How**
+
+Mark the list as a real list with correct HTML mark up, rather than creating a fake list (with only visual styling) using elements such as `<br>` or `<p>`.
```
shaundon, line 310:
```suggestion
Mark the list as a real list with correct HTML markup, rather than creating a fake list (with only visual styling) using elements such as `<br>` or `<p>`.
```

docs/src/static-pages/accessibility.md
```diff
@@ -383,21 +434,39 @@ Some people may struggle to see content unless they are able to magnify it.
...
+  - "Sticky" UI can make parts of the screen inaccessible, especially on smaller screens
 
 ### **Tests**
 - Magnify content up to 400% and check that all elements are still visible and operational across all devices and all screen sizes
 
+## Small screen size
+
+### **Why**
+
+Some people use mobiles with screens as small as `320px` wide
```
shaundon, line 456:
```suggestion
Some people use devices with screens as small as `320px` wide
```

docs/src/static-pages/accessibility.md
```diff
@@ -407,18 +476,19 @@ Some people may struggle to see content unless they are able to magnify it.
...
+
 Respect the system setting for "reduce motion". Use softer transitions, don't auto-advance content, and pause animations/videos by default.
 
-### **Do / Don't**
-- Do use soft transitions
-- Don't use high motion transitions
-- Do allow users to control motion themselves
+**Web**: Use 'Reduce Motion Media Query'
+
+**App**: System AI we can query
```
shaundon, line 491:
```suggestion
**App**: System API we can query
```

docs/src/static-pages/accessibility.md
```diff
@@ -430,11 +500,13 @@ Respect the system setting for "reduce motion". Use softer transitions, don't au
 
 # ARIA labels and patterns
 
-## ARIA labels
+## ARIA labels (web)
+
+(On native, the equivalent of an aria-label is an `accessibilityLabel`, and the equivalent of an aria-role is an `accessibilityTrait`)
```
shaundon, line 505:
```suggestion
(On native, the equivalent of an aria-label is an `accessibilityLabel`, and the equivalent of an aria-role is an `accessibilityTrait`).
```

shaundon (changes requested):

docs/src/static-pages/accessibility.md
olliecurtis: I know this is correct, we should probably get it updated on the source docs so we don't deviate and get confused with the correct source of truth

shaundon (approved):

PR #828, opened 2020-11-26
[NO-JIRA] Add aria-live demo component

georgegillams:

Screenshot_2020-11-26 ARIA Live Backpack

Screenshot_2020-11-26 Chip Backpack

Screenshot_2020-11-26 Banner alert Backpack


shaundon (dismissed):
Technically `preamble` is one word, not two, so the propType shouldn't be `preAmble` but it probably doesn't matter

shaundon (approved):

PR #836, opened 2020-12-07
[BPK-4315]Add ability to create toc for mdx files

olliecurtis:
This PR adds the ability to automatically generate a Table of Contents for mdx files. 

It does this by using the [`remark-toc`](https://github.com/remarkjs/remark-toc) plugin for mdx-loader.

To do this we add the `toc` plugin to the webpack config and in this PR I set it to some sensible defaults
- `tight: true` - This complies the list tightly, by default it was false which produced some weird spacing issues so this makes it fit more in with out style
- `maxDepth: 2` - This sets the max heading depth that a toc will be created by, I chose 2 as most of our pages use headings 2 for each of the sections, and when it came to the dev guidelines page created a massive list with - Why, Who, How, etc which made the list massive and unncessary.

With the above settings it produces a table like the following:
Screenshot 2020-12-07 at 14 53 40

or a more complex version:
Screenshot 2020-12-07 at 14 54 03


docs/src/layouts/links.js
```diff
@@ -487,11 +487,13 @@ export default [
   {
     id: 'USING_BACKPACK',
     category: 'Using Backpack',
+    sort: true,
```
olliecurtis, line 490: I changed these as it was bugging me it wasn't in alphabetical order (please don't be mad 😆 )

docs/src/static-pages/accessibility.mdx
```diff
@@ -4,6 +4,8 @@ subtitle: To ensure that we are building the best possible products that work fo
 category: guidelines
 ---
 
+# Table of Contents
```
olliecurtis, line 7: As there are multiple sections in this page with large grouped sections - we only put one `#` here so we can get the major h1 headings (We could always make the h1s to h2s, and go down a size if we prefer?)

docs/src/static-pages/website-contribution-guide.mdx
```diff
@@ -68,6 +70,10 @@ Your file must begin with a small block of text that includes some details about
 
 Underneath the page information block, you can now add your markdown content.
 
+Should your new page require a table of contents, after the information block created above just add the following and once the page goes live a table of contents will be automatically created:
```
olliecurtis, line 73: Updating to relfect the guide of if people want to add a ToC they can - we decided to make this optional as some pages if they are just images e.g. photography - they don't need a ToC

docs/src/layouts/links.js
shaundon: omg how could you 😂

shaundon (approved):
Looks great! I think your sensible defaults are perfect too.

I haven't run it locally, I'm going off the code and screenshots.

georgegillams:
Is `## Table of Contents` a default `remark-toc` thing?

olliecurtis:
@georgegillams - 

> Is `## Table of Contents` a default `remark-toc` thing?

Yes indeed! you could also use `toc`, or `table-of-contents` see here for more info - https://github.com/remarkjs/remark-toc#remarkusetoc-options

georgegillams:
> @georgegillams -
> 
> > Is `## Table of Contents` a default `remark-toc` thing?
> 
> Yes indeed! you could also use `toc`, or `table-of-contents` see here for more info - [remarkjs/remark-toc#remarkusetoc-options](https://github.com/remarkjs/remark-toc#remarkusetoc-options)

RTFM, got it 😂

georgegillams (approved):

PR #837, opened 2020-12-09
[BPK-4326] Remove old redirects

shaundon:
Simplifying things by removing old redirects. Most of them are from years and years ago.


docs/src/constants/redirect-routes.js
```diff
@@ -20,124 +20,5 @@ import * as ROUTES from './routes';
...
-  '/components/web/tables': ROUTES.TABLE,
-  '/components/web/text': ROUTES.TEXT,
-  '/components/web/tickets': ROUTES.TICKET,
-  '/components/web/tooltips': ROUTES.TOOLTIP,
-  '/components/web/typography': ROUTES.TEXT,
-  '/tokens/layout': ROUTES.SPACINGS,
-  '/tokens/colors': ROUTES.COLORS,
-  '/using/contributing': ROUTES.USING_BACKPACK,
-  '/style-guide/copywriting': ROUTES.GENERATED_GUIDELINES_COPYWRITING,
   '/style-guide/tone-of-voice': ROUTES.TONE_OF_VOICE,
```
shaundon, line 23: I left one in to aid self documentation, so when we need to add new ones in future it's clear how they work.

olliecurtis (approved):

PR #838, opened 2020-12-09
[BPK-4327] Replace GitHub link with icons link

shaundon:
The GitHub link sucks, it only goes to the web repo and we now have direct links to the right part of GitHub on each component page.

A lot of people come to our website for icons, so I replaced it with a link to that instead. I put it at the top as I'm confident that components and icons are the top two things people come to us for.

Screen Shot 2020-12-09 at 13 25 24


olliecurtis:
> I put it at the top as I'm confident that components and icons are the top two things people come to us for.

To confirm this confidence to you can check the analytics for the page accesses to see what is most popular, as the original ordering was based on this 😄

olliecurtis (dismissed):

shaundon:
> To confirm this confidence to you can check the analytics for the page accesses to see what is most popular, as the original ordering was based on this 😄

True! I'll have a look

shaundon:
Funnily enough, icons is our 12th most visited page. But I reckon that's a symptom of the problem because people struggle to find them, based on the amount of icon-related queries we get

PR #843, opened 2020-12-10
[BPK-4329] Set meta tags on individual pages

shaundon:
How it looks in practice:

## With a regular, non-MDX page
Screen Shot 2020-12-10 at 15 07 59

## With an MDX page that doesn't have a subtitle
_Note: I did try and get it to show the platform but I couldn't get it to work, it would always say Android when running from `dist`. I gave up eventually._
Screen Shot 2020-12-10 at 16 06 44

## With an MDX page that has a subtitle
Screen Shot 2020-12-10 at 16 07 00


docs/src/components/MarkdownPage/MarkdownPage.js
```diff
@@ -32,6 +32,17 @@ import STYLES from './MarkdownPage.scss';
 
 const getClassName = cssModules(STYLES);
 
+const SEOElements = (props: { description: ?string, title: ?string }) => {
+  const { description, title } = props;
+  return (
+    <Helmet>
+      {title && <title>{title}</title>}
+      {description && <meta name="description" content={description} />}
+      {description && <meta name="og:description" content={description} />}
```
shaundon, line 41: I tried doing this:

```jsx
{description && (
 <>
    <meta name="description" content={description} />
    <meta name="og:description" content={description} />
 </>
)}
```

..but it caused an error in `Helmet`, it really didn't like this format. So I went for this way instead.

georgegillams (approved):

docs/src/components/MarkdownPage/MarkdownPage.js
```diff
@@ -32,6 +32,17 @@ import STYLES from './MarkdownPage.scss';
 
 const getClassName = cssModules(STYLES);
 
+const SEOElements = (props: { description: ?string, title: ?string }) => {
+  const { description, title } = props;
+  return (
+    <Helmet>
+      {title && <title>{title}</title>}
```
georgegillams, line 39: Helmet has a dedicated prop for `title` as it can do some magic with the title template.

```suggestion
    <Helmet title={title}>
```

georgegillams (changes requested):

docs/src/components/MarkdownPage/MarkdownPage.js
shaundon: Could you be more specific?

shaundon: Helmet's docs imply this is the best way to do it

Screen Shot 2020-12-10 at 08 56 20

georgegillams (dismissed):

docs/src/components/MarkdownPage/MarkdownPage.js
```diff
@@ -27,17 +27,31 @@ import Heading from '../Heading';
...
+      {description && <meta name="og:description" content={description} />}
+    </Helmet>
+  );
+};
+
 type Props = {
   content: any,
   fileName: string,
-  title: ?string,
+  title: string,
```
shaundon, line 52: Title is no longer optional, as I've ensured all MDX pages have one.

docs/src/layouts/DefaultLayout/DefaultLayout.js
```diff
@@ -43,7 +43,10 @@ const DefaultLayout = props => {
     >
       <div id="pagewrap">
         <Analytics />
-        <Helmet titleTemplate="%s | Backpack" />
+        <Helmet
+          titleTemplate="%s | Backpack"
+          defaultTitle="Backpack — Skyscanner's Design System"
```
shaundon, line 48: Using this prop as a fallback when the page doesn't have a title.

docs/src/pages/Home/HomePage.js
```diff
@@ -143,22 +142,19 @@ const HomePage = (props, context) => {
 
   return (
     <section>
-      <Helmet title="Backpack" />
```
shaundon, line 146: Using the fallback defined above instead. This looks better than the previous value of `Backpack | Backpack`.

docs/src/pages/Home/HomePage.js
```diff
@@ -143,22 +142,19 @@ const HomePage = (props, context) => {
...
       <div
         style={{ backgroundImage: `url(${heroImage.image})` }}
         className={getClassName('bpkdocs-home-page__hero')}
       >
         <div className={getClassName('bpkdocs-home-page__hero-logo-container')}>
           <div style={{ flex: 1 }}>
-            {/* eslint-disable backpack/use-components */}
```
shaundon, line 153: I was getting an error because this wasn't used, so I removed it.

docs/src/components/MarkdownPage/MarkdownPage.js
```diff
@@ -53,16 +67,22 @@ const MarkdownPage = (props: Props) => {
     title,
   } = props;
 
+  /*
+  Currently we only want to include the page head for non-component
+  pages. Only component pages have platform metadata, so this is a
+  simple way to check for it. In the future we may wish to change this
+  logic.
+  */
+  const includePageHead = title && !platform;
```
georgegillams, line 76: Could you explain why we don't want to use this for all MD pages?

georgegillams: Is it because they're subpages?

docs/src/components/MarkdownPage/MarkdownPage.js
```diff
@@ -27,17 +27,31 @@ import Heading from '../Heading';
...
 const getClassName = cssModules(STYLES);
 
+const SEOElements = (props: { subtitle: ?string, title: string }) => {
+  const { subtitle, title } = props;
+  const description =
+    subtitle || `${title} — Backpack, Skyscanner's design system`;
+  return (
+    <Helmet>
+      <title>{title}</title>
+      <meta property="twitter:title" content={`${title} — Backpack`} />
```
georgegillams, line 42:
```suggestion
      <meta property="twitter:title" content={`${title} | Backpack`} />
```

docs/src/components/MarkdownPage/MarkdownPage.js
shaundon: Ah, sure. It's because component pages are wrapped in the platform tab switcher, so they don't need the page head. Do you think I should update the comment to make this clear?

Prior to this PR we were still doing this but the logic was just `title` and didn't involve platform. The reason I changed it in this PR was because I added titles to the Android/iOS/RN text MDX pages that could be used in the metadata.

georgegillams: Don't we have a prop called `wrapped` or something which would be much more self-documenting?

shaundon: I thought about doing this, but I think that's actually a worse long-term experience. We'd be tying in presentational logic in the MDX files, because we'd need to do this:

```yaml
---
title: Text
platform: ios
wrapped: true
---
```

The MDX files shouldn't have to be aware of how they're presented, they should only contain stuff related to the content. 

Let's say in future we decide to stop doing this page wrapping and present each component as its own page. With a `wrapped` prop we'd need to change every mdx file to remove it, whereas with my solution we would just change the logic we use in one place. So this feels like a better solution to me.

georgegillams: I was thinking about these pages. I guess it would be nice to avoid having to put `wrapped` on each one, but I also don't like inferring that a page is wrapped because the platform is specified 🤔

Untitled

georgegillams: I guess in this case though, `platform` is effectively `subpageName`. So if it has a `subpageName` then we can assume it doesn't need its own meta...

shaundon: Ah, that wrapped prop is actually unused, it's a holdover from before. It has no effect

docs/src/template.js
```diff
@@ -30,21 +30,27 @@ export default ({ head = {}, html = '', assets = {} }) => `<!doctype html>
...
+  <meta property="og:site_name" content="Backpack">
+  <meta property="og:title" content="Backpack">
+  <meta property="og:type" content="website">
+  <meta property="og:url" content="https://backpack.github.io">
+
   <meta property="twitter:card" content="summary">
+  <meta property="twitter:creator" content="@skyscanner">
   <meta property="twitter:image" content="https://backpack.github.io/social-preview.png">
-  <meta property="twitter:title" content="Backpack">
+  <meta name="twitter:image:alt" content="Backpack — Skyscanner's design system">
```
georgegillams, line 48: Should this be `name`?

shaundon: Ah, yeah this is a really good point. I copypasta'd from here: https://css-tricks.com/essential-meta-tags-social-media/

Screen Shot 2020-12-11 at 10 10 10

Having said that, we know the ones we have work fine with `property` so I'll change this one too, for consistency.

georgegillams (approved):

PR #851, opened 2020-12-14
[BPK-4330] Fix warnings

shaundon:
Initially I set out to remove **all** warnings, but some of them are non trivial to fix, and I think it would be a waste of effort when a lot of these files are about to become deprecated and removed in favour of MDX.


docs/src/components/ApidocLink/ApidocLink.js
```diff
@@ -105,13 +105,11 @@ export const createFromType = (type, defProps) => {
...
   <aside className={getClassName('bpkdocs-apidoc-link')}>
     {logo && (
-      /* eslint-disable backpack/use-components */
       <img
         className={getClassName('bpkdocs-apidoc-link__logo')}
         src={`/${logo.src}`}
         alt={logo.alt}
       />
-      /* eslint-enable backpack/use-components */
     )}
```
georgegillams, line 113: Shouldn't we just make this a BPKImage? We know the dimensions so should be possible?

shaundon: 😉 

Screen Shot 2020-12-14 at 14 22 30

docs/src/components/Typography/font-face.scss
```diff
@@ -26,8 +28,8 @@
 
   &__text {
     width: max-content;
-    font-size: $bpk-spacing-base * 2;
-    line-height: $bpk-spacing-base * 2;
+    font-size: $size;
+    line-height: $size;
```
georgegillams, line 32: Was this producing a warning/error before? Or did you just refactor for fun?

docs/src/components/ApidocLink/ApidocLink.js
georgegillams: 🤦

shaundon: I've decided to not update any components that we expect to remove soon in favour of MDX, so technically we should but I'm going to leave it.

docs/src/components/Typography/font-face.scss
shaundon: Yeah it was producing a weird stylelint warning

georgegillams (approved):

PR #854, opened 2020-12-14
[BPK-4316] Migrate animation page to MDX

shaundon:
In this PR I've only migrated one page to validate the approach I'm taking. Once approved, I'll migrate the rest of the token pages.

Before: https://backpack.github.io/tokens/animation
After: https://backpack.github.io/prs/854/tokens/animation (once CI publishes it)

I made a new `<Tokens />` component to handle displaying the tokens. I reused as much of `<TokenTable />` as I could, but I couldn't reuse it completely because it does all the platform switching stuff. Also it's so hard to understand wtf is happening with that stuff, the code is not readable at all. We can delete it once we move all of these.


docs/src/components/Tokens/Tokens.js
```diff
@@ -0,0 +1,90 @@
...
+          <BpkTableHeadCell alternate>Value</BpkTableHeadCell>
+        </BpkTableRow>
+      </BpkTableHead>
+      <BpkTableBody>
+        {!formattedAndFilteredtokens.length && (
+          <BpkTableRow key="notAvailable">
+            <BpkTableCell colSpan="2">N/A</BpkTableCell>
+          </BpkTableRow>
+        )}
+        {!!formattedAndFilteredtokens.length &&
```
olliecurtis, line 76: Is this supposed to be `!!`?

olliecurtis (commented):
LGTM! Once the PR deploys and can see it running will approve

docs/src/components/Tokens/Tokens.js
shaundon: It's something I copied from `TokensTable` but I think I can just remove it.

olliecurtis (approved):
Checked PR build and all seems to be working fine and look as expected

docs/src/components/Tokens/Tokens.js
georgegillams: That's essentially casting it from a number to a boolean by inverting it twice.

shaundon: I know, but it's not needed, right? As `1` is truthy and `0` is falsy so the cast isn't necessary.

I merged it anyway but I'll attempt to remove in the next PR I make to migrate the other pages.

georgegillams: It is, because otherwise you'll end up with an extraneous number in the dom.

https://codesandbox.io/s/unruffled-moon-9um7p?file=/src/App.js

georgegillams: `true && <...>` -> `<...>`
`false && <...>` -> null
`5 && <...>` -> `<...>`
`0 && <...>` -> 0

shaundon: Ahhh ok

PR #857, opened 2020-12-15
[No Jira] Add PR links to PR builds

shaundon:


docs/src/components/TopBanner/TopBanner.scss
```diff
@@ -23,5 +23,11 @@
   justify-content: space-between;
   background-color: $bpk-color-kolkata;
   color: $bpk-color-sky-blue-shade-03;
+  font-weight: bold;
   text-align: center;
+
+  &__link {
+    margin-left: $bpk-spacing-base;
+    color: $bpk-color-abisko;
```
shaundon, line 31: Abisko is more accessible on kolkata.

docs/src/components/TopBanner/TopBanner.js
```diff
@@ -17,17 +17,36 @@
...
+        {links.map(link => (
+          <BpkLink
+            href={link.href}
+            className={getClassName('bpkdocs-top-banner__link')}
+          >
+            {link.title}
+          </BpkLink>
+        ))}
+      </div>
+    </div>
```
shaundon, line 49: Refactored this component to take a `links` array. This makes it easier to add a class to each link. I didn't want to do it at the point of use in `index.js` because then we'd need to start importing styles into `index.js`, which felt a bit wrong.

docs/src/components/TopBanner/TopBanner.js
```diff
@@ -17,17 +17,36 @@
...
+  className: ?string,
+  links: Array<{ title: string, href: string }>,
+};
+
+const TopBanner = (props: Props) => {
+  const { children, className, links, ...rest } = props;
   return (
-    <div {...rest} className={getClassName('bpkdocs-top-banner', className)} />
+    <div {...rest} className={getClassName('bpkdocs-top-banner', className)}>
+      {children}
```
georgegillams, line 38: If we're passing links in as an object, do we need `children`?

shaundon: `children` is used for the rest of the text — in this case it says "Pull request build"

georgegillams (approved):

PR #871, opened 2020-12-18
[NO JIRA]: Migrate overlay pages

olliecurtis:
Migrating overlay pages to markdown


shaundon (approved):
Looking good! Do you want to remove `NativeComponentPage.js` as part of this too, or separately?

olliecurtis:
> Looking good! Do you want to remove `NativeComponentPage.js` as part of this too, or separately?

Its coming in a separate PR as there is a bigger change as part of it 😄

PR #872, opened 2020-12-18
[NO JIRA]: Remove native component page

olliecurtis:
Now we have migrated all the native component pages to MDX we can remove this component.

I also fixed a devtools error with the component screenshots where it was expecting width and height but we are now only using aspect ratio. Which is a requirement/favourable in `BpkImage`


shaundon (approved):
Nice!

PR #873, opened 2020-12-18
Remove docs page builder

olliecurtis:
This PR finally does it! DocsPageBuilder has officially been removed!

Changes:
- Removed DocsPageBuilder
- Moved components that were bundled in the PageBuilder folder into their own component pages as they are still in use and provide value
- Updated file reference paths
- Migrated the three pages using the DocsPageBuilder to alternative solutions


shaundon (commented):
Looks great from the diff, I just wanna check the built site before approving

shaundon (approved):

PR #889, opened 2021-01-29
[BPK-4372] Migrate to simpler web accordion page

shaundon:
* Added support for markdown pages representing web components
* Removed old web accordion page and replaced it with a markdown page that uses the component examples from the web repo, reducing duplication


docs/src/components/MarkdownPage/AdditionalLinks.js
```diff
@@ -99,6 +99,20 @@ const AdditionalLinks = (props: Props) => {
           </BpkLink>
         )}
 
+        {/* Web npm link */}
+        {platform && platform === PLATFORMS.web && githubPath && (
```
shaundon, line 103: To reduce duplication, we use the `githubPath` metadata item for the NPM link, because we know they match on web components.

docs/src/components/PresentationBlock/PresentationBlock.js
```diff
@@ -26,35 +26,31 @@ const getClassName = cssModules(STYLES);
 
 const PresentationBlock = props => {
   const classNames = [getClassName('bpkdocs-presentation-block')];
-  const { className, darkBackground, whiteBackground, ...rest } = props;
```
shaundon, line 29: Most components were setting `whiteBackground`, so I decided to make it the default behaviour to reduce the need to pass it in.

georgegillams (approved):
It's perfect

PR #890, opened 2021-01-29
Add 'edit page' buttons to all MDX component pages

shaundon:


docs/src/components/MarkdownPage/AdditionalLinks.js
```diff
@@ -203,7 +203,11 @@ const AdditionalLinks = (props: Props) => {
           secondary
           blank
           className={getClassName('bpkdocs-markdown-page__edit-button')}
-          href={`https://github.com/skyscanner/backpack-docs/edit/master/docs/src/static-pages/${fileName}`}
+          href={
+            platform
+              ? `https://github.com/skyscanner/backpack-docs/edit/master/docs/src/pages/components/${fileName}`
```
shaundon, line 208: If/when we move the mdx files to live within their respective repos, this logic will become more complex by looking at the platform, but yagni for now

k0nserv (commented):
Wouldn't it be better to derive `fileName` automatically from the path if it will be predictable in 99% of cases. We can still allow setting it explicitly, but it seems like a lot of repettition for the obvious cases?

olliecurtis (dismissed):

shaundon:
> Wouldn't it be better to derive fileName automatically from the path if it will be predictable in 99% of cases. We can still allow setting it explicitly, but it seems like a lot of repettition for the obvious cases?

I might be being dense but I couldn't figure out how we'd do this at the time. Maybe I could use a Node API to get the current working directory and file name?

k0nserv:
[`path.relative`](https://nodejs.org/api/path.html#path_path_relative_from_to) should do the trick

image

shaundon:
@k0nserv you're a freaking genius, I replaced it all with something derived automatically from Webpack, now instead of 87 files changed it's 5!

k0nserv (approved):

PR #899, opened 2021-02-11
[BPK-4401]: Migrate to main branch

olliecurtis:


lerna.json
```diff
@@ -7,7 +7,7 @@
     },
     "publish": {
       "ignoreChanges": ["bpk-tether", "package-lock.json"],
-      "allowBranch": "master"
+      "allowBranch": "main"
```
olliecurtis, line 10: Saw this file and wasn't sure if we still need it or not?

As we don't publish anything from here so think it was left over from when we migrated the repo?

georgegillams: Seems redundant

georgegillams (approved):

PR #910, opened 2021-02-24
[BPK-3615]: Add accessibility Dos and Don'ts for BpkLink

olliecurtis:
+ [ ] Check this if you have read and followed the [contributing guidelines](https://github.com/Skyscanner/backpack/blob/main/CONTRIBUTING.md)

### DIFF
--- a/docs/src/pages/components/Link/WebLinksPage.js
+++ b/docs/src/pages/components/Link/WebLinksPage.js
@@ -64,6 +64,18 @@ const LinkPage = () => (
     readme={linkReadme}
     examples={components}
     packageName="bpk-component-link"
+    usageTable={{
+      dos: [
+        'Use an "aria-label" for each link on a page should clearly describe where it will take you if you choose to follow it. E.g. "A link to the Skyscanner privacy policy."',
+        'If links open in a new window, users should be informed of this.',
+        'If a link contains a document ensure users are aware of this (type and size of file).',
+        'CSS can be used to hide text to hide a portion of the link text to provide more information to users of screen readers.',
+        'Assigning aria-describedby and an ID can be used to associate text that is already present on the page with the link.',
+      ],
+      donts: [
+        "Don't just use the link text as this may not be descriptive. E.g 'privacy policy'.",
+      ],
+    }}
   />
 );
 
--- a/docs/src/pages/components/Link/WebLinksPage.js
+++ b/docs/src/pages/components/Link/WebLinksPage.js
@@ -66,14 +66,15 @@ const LinkPage = () => (
     packageName="bpk-component-link"
     usageTable={{
       dos: [
-        'Use an "aria-label" for each link on a page should clearly describe where it will take you if you choose to follow it. E.g. "A link to the Skyscanner privacy policy."',
-        'If links open in a new window, users should be informed of this.',
-        'If a link contains a document ensure users are aware of this (type and size of file).',
+        'Each link on a page should clearly describe where it will take you if you choose to follow it. E.g. "Skyscanner privacy policy."',
+        'If a link contains an element such as an icon you should ensure it has suitable alt text for assistive technologies.',
+        'If links open in a new window, users of assitive technologies should be informed of this by applying the appropriate aria-label. E.g. The aria-label could be "opens in a new tab".',
+        'If a link contains a document ensure users are aware of this (type and size of file), this can either be provided in the link text or using an aria-label.',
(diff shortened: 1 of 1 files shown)


docs/src/pages/components/Link/WebLinksPage.js
```diff
@@ -64,6 +64,18 @@ const LinkPage = () => (
     readme={linkReadme}
     examples={components}
     packageName="bpk-component-link"
+    usageTable={{
+      dos: [
+        'Use an "aria-label" for each link on a page should clearly describe where it will take you if you choose to follow it. E.g. "A link to the Skyscanner privacy policy."',
```
olliecurtis, line 69: Need to change this to not mention `aria-label` as the link text itself should be descriptive

docs/src/pages/components/Link/WebLinksPage.js
```diff
@@ -64,6 +64,18 @@ const LinkPage = () => (
...
+        'If links open in a new window, users should be informed of this.',
+        'If a link contains a document ensure users are aware of this (type and size of file).',
+        'CSS can be used to hide text to hide a portion of the link text to provide more information to users of screen readers.',
+        'Assigning aria-describedby and an ID can be used to associate text that is already present on the page with the link.',
+      ],
+      donts: [
+        "Don't just use the link text as this may not be descriptive. E.g 'privacy policy'.",
```
olliecurtis, line 76: Because of the above comment this Don't is probably not needed

docs/src/pages/components/Link/WebLinksPage.js
```diff
@@ -64,6 +64,18 @@ const LinkPage = () => (
     readme={linkReadme}
     examples={components}
     packageName="bpk-component-link"
+    usageTable={{
+      dos: [
+        'Use an "aria-label" for each link on a page should clearly describe where it will take you if you choose to follow it. E.g. "A link to the Skyscanner privacy policy."',
+        'If links open in a new window, users should be informed of this.',
+        'If a link contains a document ensure users are aware of this (type and size of file).',
+        'CSS can be used to hide text to hide a portion of the link text to provide more information to users of screen readers.',
```
georgegillams, line 72: I don't know if this is so relevant to our link. Does it accept any array of children or only text?

docs/src/pages/components/Link/WebLinksPage.js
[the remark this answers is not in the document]
georgegillams: I think it's worth saying something like "don't use an aria-label if the link text is already completely satisfactory. Aria-label should be a last resort"

georgegillams (changes requested):

docs/src/pages/components/Link/WebLinksPage.js
georgegillams: I guess there's also the point that if the link only contains an icon, that icon needs to have a suitable `alt` text to ensure it works for AT (and users on low bandwidth which might prevent the icon showing)

docs/src/pages/components/Link/WebLinksPage.js
olliecurtis: It supports an array of children or text depending on what the consumer passes - which I guess is for if an icon is passed to it

docs/src/pages/components/Link/WebLinksPage.js
```diff
@@ -64,6 +64,19 @@ const LinkPage = () => (
     readme={linkReadme}
     examples={components}
     packageName="bpk-component-link"
+    usageTable={{
+      dos: [
+        'Each link on a page should clearly describe where it will take you if you choose to follow it. E.g. "Skyscanner privacy policy."',
+        'If a link contains an element such as an icon you should ensure it has suitable alt text for assistive technologies.',
+        'If links open in a new window, users of assitive technologies should be informed of this by applying the appropriate aria-label. E.g. The aria-label could be "opens in a new tab".',
```
georgegillams, line 71: For these two do we need to make it clearer that the aria label should include the link text + the extra info? Otherwise someone might make a screen-reader just read "opens in a new tab" 🤔

georgegillams (commented):
This is looking great now 👍 
Just one small question remaining from me

docs/src/pages/components/Link/WebLinksPage.js
olliecurtis: Yeah for sure! Wasn't 100% which order it would read it out so if it does the aria label first then the actual content, then totally makes sense to include both - will update this now 😄

georgegillams (approved):

PR #923, opened 2021-03-11
[BPK-3607] Migrate tooltips to new docs format

shaundon:


olliecurtis (commented):
LGTM! Will check it once deployed

PR #924, opened 2021-03-11
[BPK-4389]: Migrate two components to new stories

olliecurtis:
Migrate Horizontal Nav to new format

+ [x] Check this if you have read and followed the [contributing guidelines](https://github.com/Skyscanner/backpack/blob/main/CONTRIBUTING.md)


docs/src/pages/components/HorizontalNav/WebHorizontalNav.mdx
```diff
@@ -0,0 +1,54 @@
...
+<PresentationBlock>
+  <NotUnderlinedExample />
+</PresentationBlock>
+
+## Light appearance
+
+An alternate appearance is available for display over dark backgrounds.
+
+<PresentationBlock style={{ backgroundColor: colorSkyBlueShade02 }}>
+  <LightAppearanceExample />
```
shaundon, line 49: We have `BpkDarkExampleWrapper` in the main repo for this, the idea is that the docs site should never have to be concerned about this presentation and everything is built into the example itself. 

In this case I'd recommend leaving the inline style in _for now_, but then in your next round of PRs, change `LightAppearanceExample` in `bpk-horizontal-nav/examples.js` to use `BpkDarkExampleWrapper`, then come back and remove the inline style from here as it will then not be needed.

shaundon (approved):

docs/src/pages/components/HorizontalNav/WebHorizontalNav.mdx
olliecurtis: Yes the example is using this in the main repo but the container was not big enough for it to properly display which is why I have had to do this so that it could be shown off properly

olliecurtis: This is what it looks like without the inline style

Screenshot 2021-03-12 at 10 47 35

shaundon: Ah I see. I think we should still fix it in the web repo though, so this page can remain ignorant of style concerns. We could just add some bottom padding on the example in the web repo to make the issue go away

georgegillams: `<PresentationBlock` has a prop for exactly this 🤔

georgegillams: https://github.com/Skyscanner/backpack-docs/pull/925

PR #931, opened 2021-03-19
[BPK-4435] Remove code (phase 1)

shaundon:
There'll be more PRs after this but I'm splitting it into phases.

It just removes a bunch of files and in a couple of cases migrates to using simpler syntax so that a bespoke component can be deleted.


shaundon:
CI has failed for non-code reasons, so I think it's fine.

georgegillams (approved):

PR #933, opened 2021-03-19
[BPK-4435] Remove more code

shaundon:
Prior to this PR, we had two ways of rendering Markdown:

1. `BpkMarkdownRenderer`, used for rendering readmes, using `react-markdown` behind the scenes
2. `MDX`, used for everything else

This PR removes `BpkMarkdownRenderer` and all associated stuff so we can unify on one solution. It does the following:

* Changes the Webpack config so that `.md` files are run through `MDX` instead of `raw-loader` (and removes the no-longer-used `raw-loader` dependency)
* Makes use of MDX's transclusion to directly include readmes instead of having to wrap them in a component
* Removes components that are no longer used, like `<Readme>` and `<BpkMarkdownRenderer>`.

There is one downside to this. Because we're not transforming readmes anymore, they're placed on the page exactly as written. Previously we were using some logic to remove the initial `H1` and `blockquote` and increase each heading by one increment. I investigated doing this through Webpack config now instead, but it's non trivial and I don't think it's really a problem that the readmes are presented as written. 

They still look fine, and I think the simplicity win is worth the slight degradation. If you don't agree, we should either fix this now or do it as a follow up later.


docs/src/pages/components/StarRating/WebStarRating.mdx
```diff
@@ -4,10 +4,10 @@ platform: web
 githubPath: bpk-component-star-rating
 ---
 
-import Readme from 'components/Readme';
+
 import PresentationBlock from 'components/PresentationBlock';
 import UsageTable from 'components/UsageTable';
-import readme from 'bpk-component-ticket/README.md';
+import Readme from 'bpk-component-star-rating/README.md';
```
shaundon, line 10: The wrong readme was being imported.

docs/src/pages/components/Images/WebImages.mdx
```diff
@@ -4,19 +4,19 @@ platform: web
...
 import PresentationBlock from 'components/PresentationBlock';
 
-import { 
-  DefaultExample, 
-  RoundedCornersExample, 
-  FullWidthExample, 
-  WithAnimationExample, 
-  WithLazyLoadingExample, 
-  WithLazyLoadingAndAnimationExample, 
-  BackgroundImageExample
```
georgegillams, line 17: We should really introduce some basic prettier rules for this. It already supports MDX so shouldn't be too hard.

georgegillams (approved):

docs/src/pages/components/Images/WebImages.mdx
shaundon: Good idea, I'll make a card for it.

PR #978, opened 2021-04-27
Use github API to check if PR is external

marianeum:
Updated the danger check to use the github API to check if a PR is external, rather than using a manually maintained list.

+ [x] Check this if you have read and followed the [contributing guidelines](https://github.com/Skyscanner/backpack/blob/main/CONTRIBUTING.md)


dangerfile.js
```diff
@@ -25,17 +25,15 @@ import fs from 'fs';
...
   { word: 'react native', reason: 'Please use React Native with capitals' },
 ];
 
-const BACKPACK_SQUAD_MEMBERS = meta.maintainers.map(
-  maintainer => maintainer.github,
-);
 const author = danger.github.pr.user.login;
-const isPrExternal = !BACKPACK_SQUAD_MEMBERS.includes(author);
+const isPrExternal = !danger.github.api.orgs.checkMembershipForUser({
+  org: 'backpack',
```
olliecurtis, line 34: We should probably be using the org `Skyscanner` as that is where the repo and its users sit rather than a specific Backpack org. We have one but that holds different things and not connected to running danger here


```suggestion
  org: 'skyscanner',
```

olliecurtis (approved):

olliecurtis (changes requested):
Sorry clicked the wrong button, see above

dangerfile.js
marianeum: Ah if we want to use Skyscanner it might be even easier :) Right now the list of maintainers matched the backpack org so I used that. We just use it for the thank you message, which might be nice for non-backpack Skyscanner employees as well?

olliecurtis: Yeah it would be great if we can allow it to run for any Skyscanner employees contributing changes, so they can get the danger feedback as well 😄

olliecurtis: With these changes too it might also mean we can completely remove the `meta.json` file as it may not be used outside of this script, and the npm sections were left over from when we split the docs from the Backpack web repo

marianeum: Yep danger runs for everybody making contributions - the `isPrExternal` check just adds a thank you message for anybody external, so if we use the backpack org we won't get the thank you, but everybody else will

olliecurtis: Ah yes I see now its literally just used to add that message! Am wondering if that is needed at all and we just post the message regardless of who it is? 😆

marianeum: Haha maybe! On the android project we also have a warning if it's done as a fork that CI will fail - not sure if other projects also have more checks. It did feel weird, though getting a thank you for essentially doing my job

olliecurtis: In our other projects we put checks in place that don't run CI steps that are dependent on things only available on the main repo (e.g. tokens for CI) so we don't give that warning as CI won't run in parts

shaundon: Just for some context, I think we did used to use it for more than just the thank you message but that need went away, so now it's only used for this trivial lil' thing

meta.json
```diff
@@ -1,26 +1,16 @@
 {
```
olliecurtis, line 1: Now #979 has been merged and the npm users have been removed, we can just delete this file now 🎉

olliecurtis (approved):

meta.json
marianeum: done!

dangerfile.js
```diff
@@ -38,11 +40,37 @@ const isPrExternal = !danger.github.api.orgs.checkMembership({
...
+  'https://media.giphy.com/media/xULW8v7LtZrgcaGvC0/giphy.gif', // Dog
+  'https://media.giphy.com/media/GghJ32T5oPR8Q/giphy.gif', // Leslie Knope
+  'https://media.giphy.com/media/26AHAw0aMmWwRI4Hm/giphy.gif', // David Mitchell
+  'https://media.giphy.com/media/mbhseRYedlG5W/giphy.gif', // That guy from Who's Line Is It Anyway who looks like Bill Murray
+  'https://media.giphy.com/media/3o6ZsXRBB9E67nUjL2/giphy.gif', // We love you
+  'https://media.giphy.com/media/l3V0sNZ0NGomeurCM/giphy.gif', // Bowie
+  'https://media.giphy.com/media/3rgXBvoeXt3MXlqhO0/giphy.gif', // Amazement
+  'https://media.giphy.com/media/1OnDp7RwgphjG/giphy.gif', // Kumamon
+  'https://media.giphy.com/media/1lk1IcVgqPLkA/giphy.gif', // Cap salute
+  'https://media.giphy.com/media/l3V0lN5bUX7AKtdW8/giphy.gif', // Moira Schitt's Creek
```
shaundon, line 57: ❤️

shaundon (approved):

shaundon (approved):

olliecurtis (approved):

PR #979, opened 2021-04-27
Removing old npm scripts not used by project

olliecurtis:
Inspired by #978 I noticed that we have reference in our `meta` file to our npm accounts which are(were) used in the `scripts/npm` folder which were left over from when the docs were split from the main Skyscanner/Backpack repo.

As such these npm scripts are not referenced anywhere in the docs repo so can be safely cleaned up from the repo as we don't publish any npm packages or projects from this repo.


marianeum (approved):
nice one!

PR #988, opened 2021-05-05
[No Jira] Replace BpkRouterLink with home-grown component

shaundon:
`BpkRouterLink` is only used by the docs site, so we've decided to remove it as a component. It's a holdover from when the docs site was in the web repo. This will mean we can remove some dependencies from the web repo and simplify the codebase.

This PR pulls the component into the docs site and removes the dependency on it, so that it's no longer used anywhere.


olliecurtis (dismissed):

docs/src/components/RouterLink/RouterLink.js
```diff
@@ -0,0 +1,58 @@
...
+  }
+
+  return (
+    <NavLink
+      className={classNames.join(' ')}
+      activeClassName={getClassName('bpkdocs-router-link--active')}
+      to={to}
+      {...rest}
+    >
+      {children}
```
georgegillams, line 43: Why doesn't this render a BpkLink instead of using custom styles?

georgegillams (changes requested):

docs/src/components/RouterLink/RouterLink.js
georgegillams: https://github.com/ReactTraining/react-router/blob/master/packages/react-router-dom/docs/api/Link.md#component-reactcomponent

georgegillams: In this case, FancyLink would be BpkLink

shaundon: It uses the `BpkLink` mixin behind the scenes. This is literally just copied from the existing component so I don't think it matters tbh

georgegillams: Dogfooding is important though, and it'll save us changing this component if we change the behaviour of `BpkLink` in the future.

shaundon: Dammit why are you always right 😂

I'll change it, thanks for the feedback!

georgegillams (approved):

PR #1006, opened 2021-06-02
[KOA-4578]: Adopting new Backpack foundations packages in docs site

olliecurtis:
This PR adopts th new foundations packages for the docs site and all the required changes to use the correct packages, where necessary.


package.json
```diff
@@ -77,6 +77,11 @@
     "@github-docs/frontmatter": "^1.3.1",
     "@mdx-js/loader": "^1.6.21",
     "@mdx-js/react": "^1.6.21",
+    "@skyscanner/bpk-android": "^1.0.0",
+    "@skyscanner/bpk-ios": "^1.0.0",
+    "@skyscanner/bpk-react-native": "^1.0.0",
```
shaundon, line 82: Do we need these three? I couldn't see them being used anywhere in this diff, just the web and svg ones

olliecurtis: Yeah they are used inside `.mdx` files - an example is here 😄 https://github.com/Skyscanner/backpack-docs/pull/1006/files#diff-5b968253cf6d7a11889eac8b1a79d6ea37c0208f3eb610588d472b4b3aa7763eR7

olliecurtis: I think a lot of them are the tokens pages where we have the tables for tokens per platforms

shaundon: Ah, I must have just missed them when reviewing the diff. Apologies!

shaundon (approved):

PR #1092, opened 2021-11-16
[no jira]: Upgrading foundations

olliecurtis:


marianeum (approved):
I noticed bpk packages are ignored in dependabot - is that intentional?

olliecurtis:
> I noticed bpk packages are ignored in dependabot - is that intentional?

@marianeum Yep, otherwise each time we release dependabot opens a million individual PRs for each package 😂 - so instead when we release Backpack web there is an automated step to make automated PRs to do the upgrade, but it might also be getting caught the scoped packages as well so may need to tweak the config to allow foundations but not web

PR #1117, opened 2021-12-15
KOA-4773 Add compose platform docs

marianeum:
In preparation for releasing compose add support for displaying compose docs. The first compose docs (Text) are available:
screencapture-0-0-0-0-8080-components-text-2021-12-15-10_44_09

The maven central link doesn't work yet, as we haven't published yet, but once we have it'll work.

### DIFF
--- a/docs/src/components/MarkdownPage/AdditionalLinks.js
+++ b/docs/src/components/MarkdownPage/AdditionalLinks.js
@@ -206,7 +206,7 @@ const AdditionalLinks = (props: Props) => {
         {/* Android compose GitHub link */}
         {platform && platform === PLATFORMS.compose && githubPath && (
           <BpkLink
-            href={`https://github.com/Skyscanner/backpack-android/tree/main/backpack-compose/src/main/java/net/skyscanner/backpack/compose/${githubPath}`}
+            href={`https://github.com/Skyscanner/backpack-android/tree/main/backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/${githubPath}`}
             blank
             className={getClassName('bpkdocs-markdown-page__link')}
           >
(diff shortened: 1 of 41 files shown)


olliecurtis (changes requested):
Code looks good to me! However checking the deployed version the maven and github links give errors when you click on them.

Seems there is no maven build for compose yet? and the github url looks to point to a dead end

Screenshot 2021-12-15 at 12 14 03

marianeum:
Yep as I mentioned above: The maven central link doesn't work yet, as we haven't published yet, but once we have it'll work.

So it points to the right package we'll be publishing on.

olliecurtis:
> Yep as I mentioned above: The maven central link doesn't work yet, as we haven't published yet, but once we have it'll work.
> 
> So it points to the right package we'll be publishing on.

ooop sorry missed that under the image 🤦 sorry about that! Does the same apply for the github link?

marianeum:
Ah sorry I missed this! That is broken indeed, thanks :)

docs/src/components/MarkdownPage/AdditionalLinks.js
```diff
@@ -174,6 +203,20 @@ const AdditionalLinks = (props: Props) => {
           </BpkLink>
         )}
 
+        {/* Android compose GitHub link */}
+        {platform && platform === PLATFORMS.compose && githubPath && (
+          <BpkLink
+            href={`https://github.com/Skyscanner/backpack-android/tree/main/backpack-compose/src/main/java/net/skyscanner/backpack/compose/${githubPath}`}
```
olliecurtis, line 209: Looks like the path is `kotlin` and not `java` in the path

```suggestion
            href={`https://github.com/Skyscanner/backpack-android/tree/main/backpack-compose/src/main/kotlin/net/skyscanner/backpack/compose/${githubPath}`}
```

olliecurtis (approved):

PR #1120, opened 2021-12-17
[KOA-4802]: Spacing table updates

olliecurtis:
Tidying up the spacing tokens page for web to remove the ugly spacing functions and their logic to just the tokens, the engineering token due to them being functions and their value.

Also in this PR I found some code we could cleanup and remove as it was not being used anywhere in the site 🎉

| Before | After |
| -- | -- |
| image | image|


docs/src/helpers/tokens-helper.js
```diff
@@ -153,57 +153,3 @@ export const tokenNameToSizeNumber = (token: {
 
   return parseInt(token.value, 10);
 };
-
-export const getTokens = (tokens: TokenList, keys: ?Array<string> = null) =>
```
olliecurtis, line 157: Removed all this code as its not being used anywhere in the docs site 🎉

anambl (approved):

PR #1177, opened 2022-04-11
[no jira]: Update Button docs to latest types and fix crash

olliecurtis:
Updating docs page to add new button types
Remove non existing outline button type (causing a page crash)
Cleaned up unused imports to reduce bundle size

Buttons Page


anambl (approved):
LGTM!

PR #1208, opened 2022-05-10
[no jira]: Fixing Web nudger page after recent component change

olliecurtis:
We recently removed the outline nudger from Backpack web however I forgot to remove this and update the docs site after removing it! 

This PR fixes that


anambl (approved):
LGTM!

PR #1306, opened 2022-07-04
[KOA-5360]: Deprecate Navigation Stack component

olliecurtis:
As we have deprecated this component in Backpack we are removing the docs pages to avoid people using it.


anambl (approved):
LGTM!

PR #1330, opened 2022-07-21
Update WebButtonPage.mdx

Fynci:
Added accessibility considerations and screen reader testing.


olliecurtis (approved):

olliecurtis:
As this is only adding text and not changing functionality we can accept the change without full CI passing due to Dean not having access to our org to fork and CI to run with secrets

PR #1480, opened 2022-10-26
Remove screenshots and badges from docs

marianeum:
Now that they're embedded in the Android docs we no longer need to include them here.


docs/src/components/MarkdownPage/AdditionalLinks.js
```diff
@@ -63,34 +63,6 @@ const AdditionalLinks = (props: Props) => {
   return (
     <div className={getClassName('bpkdocs-markdown-page__additional-links')}>
       <div>
-        {/* Android view Maven Central link */}
```
marianeum, line 66: removed this for good measure as well

frugoman (approved):

